PR #6, opened 2019-03-29
Observation intro text in docs

nielsdrost:
Added a small intro text for the USGS and GRDC datasets.


sverhoeven:
Closed during #164

PR #12, opened 2021-03-22
Fix missing values in grdc data

SarahAlidoost:
This PR replaces -999.0 values with np.nan in grdc streamflow values. Thus, other packages like hydrostats and matplotlib handle missing values correctly.

### DIFF
--- a/ewatercycle/observation/grdc.py
+++ b/ewatercycle/observation/grdc.py
@@ -116,10 +116,7 @@ def _grdc_read(grdc_station_path, start, end):
 
     # Fix missing values
     grdc_station_fix = grdc_station_select.copy()
-    if not grdc_station_fix['streamflow'].dtypes == np.float:
-        raise TypeError("Data has non-float values!")
-    else:
-        grdc_station_fix['streamflow'].replace(-999.0, np.NaN, inplace=True)
+    grdc_station_fix['streamflow'].replace(-999.0, np.NaN, inplace=True)
 
     return metadata, grdc_station_fix
 
--- a/ewatercycle/observation/grdc.py
+++ b/ewatercycle/observation/grdc.py
@@ -102,7 +102,8 @@ def _grdc_read(grdc_station_path, start, end):
         grdc_station_path,
         skiprows=header,
         delimiter=';',
-        parse_dates=['YYYY-MM-DD'])
+        parse_dates=['YYYY-MM-DD'],
+        na_values='-999')
     grdc_station_df = grdc_station_df.rename(columns={
         'YYYY-MM-DD': 'time',
         ' Value': 'streamflow'
@@ -114,11 +115,7 @@ def _grdc_read(grdc_station_path, start, end):
     # Select GRDC station data that matches the forecast results Date
     grdc_station_select = grdc_station_df.loc[start:end]
 
-    # Fix missing values
-    grdc_station_fix = grdc_station_select.copy()
-    grdc_station_fix['streamflow'].replace(-999.0, np.NaN, inplace=True)
-
-    return metadata, grdc_station_fix
+    return metadata, grdc_station_select
 
 
 def _grdc_metadata_reader(grdc_station_path, allLines):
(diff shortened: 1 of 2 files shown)


ewatercycle/observation/grdc.py
```diff
@@ -113,7 +114,14 @@ def _grdc_read(grdc_station_path, start, end):
     # Select GRDC station data that matches the forecast results Date
     grdc_station_select = grdc_station_df.loc[start:end]
 
-    return metadata, grdc_station_select
+    # Fix missing values
+    grdc_station_fix = grdc_station_select.copy()
```
sverhoeven: Why do you make a copy? Each call to `_grdc_read()` will generate a new dataframe there is no need to have 2 copies.

sverhoeven (changes requested):
This will work, but

Can you add unit tests to https://github.com/eWaterCycle/parametersetdb/blob/fix_grdc_missing_values/tests/observation/test_grdc.py
1. for the -999 to NaN replacement
2. for when streamflow is not a float,  do we have an example for this happening?

SarahAlidoost:
> This will work, but

Thanks for reviewing. 
> 
> Can you add unit tests to https://github.com/eWaterCycle/parametersetdb/blob/fix_grdc_missing_values/tests/observation/test_grdc.py
> 
> 1. for the -999 to NaN replacement

it is added now. 
> 2. for when streamflow is not a float,  do we have an example for this happening?

Right, no, we don't have. I removed the if statement that checks this.

ewatercycle/observation/grdc.py
SarahAlidoost: Before adding this copy, I got this warning from pandas:

```
miniconda3/envs/comparison/lib/python3.8/site-packages/pandas/core/series.py:4575: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  return super().replace(
```

Looking at the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy, it seems that copy() is needed after the selection in `grdc_station_df.loc[start:end]`.

sverhoeven: Ok, no warning is better. 
We could also add `na_values=-999` to [read_csv()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) on line 100 then we dont need any copy or replace.

sverhoeven (approved):
Thanks for the test. 
See inline comment for suggestion to make it even less code.

SarahAlidoost:
> Thanks for the test.
> See inline comment for suggestion to make it even less code.

Please see my response.

ewatercycle/observation/grdc.py
SarahAlidoost: Thank you for the suggestion. Using na_values in pd.read_csv(), convert -999.000 to `nan` with type `numpy.float64` that is not equal to `np.nan` anymore. Thus, the test will break. I don't know how to represent the pandas `nan` value in the test where  `{'streamflow': ('time', [123., 456.,  np.NaN(?)])}`.

sverhoeven: Hmm, a quick test works for when I use `na_values='-999',`. 

```
cat b.csv 
a,b
1,2
3,-999
```

```python
import pandas as pd
import numpy as np
z = pd.read_csv('b.csv',na_values='-999')
type(z['b'][1])
numpy.float64
np.isnan(z['b'][1])
True
```

I think it is because np.nan is a special case of a floating number. So you can just use np.nan in expected value.

When I add `na_values='-999',` to arguments read_csv on line 100 the tests/observation/test_grdc.py test passes.

SarahAlidoost: thanks, it is fixed now.

SarahAlidoost:
@sverhoeven many thanks for reviewing. Can we merge this PR?

sverhoeven (approved):
Yep, lets merge it. Code looks good and tests pass.

PR #13, opened 2021-03-23
Replace Travis CI with GitHub Actions workflow + Correct sonarcloud analysis

### LINKED ISSUE #11: Travis-ci build is failed.
After merging pull request #5, Travis-ci build is failing. Please see the last commit in this PR and the [Travis report](https://travis-ci.org/github/eWaterCycle/parametersetdb/jobs/763392107). 

It seems that Cython is a dependency for pandas, see https://pandas.pydata.org/pandas-docs/version/0.12/install.html

sverhoeven:
Fixes #11


SarahAlidoost:
@sverhoeven just to inform you that the repository is renamed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
SarahAlidoost, line 1: If I understood correctly, the workflow only install dependencies and runs pytest but not any lint!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+      types: [opened, synchronize, reopened]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
SarahAlidoost, line 17: I am wondering if we can add `3.9` to the python-version. In the [conda environment file, we defined python version as](https://github.com/eWaterCycle/ewatercycle/blob/b90f46bc23ce5aed5003e317d22f1aff3fb2f9a7/environment.yml#L7)>=3.6.  I am not sure if these two files should be consistent regarding the python version.

SarahAlidoost (approved):
@sverhoeven many thanks, I left very minor comments and approved this PR.

.github/workflows/ci.yml
sverhoeven: I took the Python version listed in https://github.com/eWaterCycle/ewatercycle/blob/dcdd5e393bf993c449a524817120cbafacfec1f7/setup.py#L39-L41
that is my source of truth for supported Python versions. 

Indeed, we should keep the versions in sync

.github/workflows/ci.yml
sverhoeven: Yep, it does the same as the Travis workflow. 
The sonar analysis will gather (linting) issues , but not make the job green.
It would be good idea to add a linter step to the job, but better to do that in another PR.

sverhoeven:
Thanks for the review.

PR #14, opened 2021-03-24
Add a conda environment file

SarahAlidoost:
This PR adds a conda environment file. (cc @bvreede )

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,12 @@
+---
+name: ewatercycle
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - python>=3.6
+  - numpy
+  - pandas
+  - pyoos
+  - ruamel.yaml
+  - xarray


nielsdrost (approved):
Thanks @SarahAlidoost !

Should we already add the rest of the dependencies, like https://github.com/eWaterCycle/era5-comparison/blob/master/environment.yml ?

SarahAlidoost:
> Thanks @SarahAlidoost !
> 
> Should we already add the rest of the dependencies, like https://github.com/eWaterCycle/era5-comparison/blob/master/environment.yml ?

Thank you for your comment, good question. I am not sure which dependencies will be eventually included in this package. I suggest adding other dependencies when they are introduced (_no strong opinion_). 

Just one issue, the `era5-comparison` environment has some packages that are used for the development, like isort, pylint, ipynb-py-convert, etc. We (with @bvreede) discussed that the the `ewatercycle` environment shouldn't include those packages. However, now I think that we need those packages because the environment file is only used by developers. When the package is published on conda-forge or pypi, users don't work with this environment file. What do you think?

nielsdrost:
Agreed to add dependencies when we need them for now.

As for dev dependencies, my current guess is that this environment should be the basis for the the lab.ewatercycle.org Jupyter kernel. So I would expect some dev dependencies.

Shall we simply merge as is for now, and add things when we need to?

SarahAlidoost:
> Agreed to add dependencies when we need them for now.
> 
> As for dev dependencies, my current guess is that this environment should be the basis for the the lab.ewatercycle.org Jupyter kernel. So I would expect some dev dependencies.

So, if I understood correctly, the lab.ewatercycle.org Jupyter kernel as defined [here](https://github.com/eWaterCycle/infra/blob/master/roles/ewatercycle/tasks/main.yml) will be merged into this environment.yml file. For the future, is it possible to keep only one of these files?

> 
> Shall we simply merge as is for now, and add things when we need to?

I am fine with merging this PR. :+1:

nielsdrost:
> > Agreed to add dependencies when we need them for now.
> > As for dev dependencies, my current guess is that this environment should be the basis for the the lab.ewatercycle.org Jupyter kernel. So I would expect some dev dependencies.
> 
> So, if I understood correctly, the lab.ewatercycle.org Jupyter kernel as defined [here](https://github.com/eWaterCycle/infra/blob/master/roles/ewatercycle/tasks/main.yml) will be merged into this environment.yml file. For the future, is it possible to keep only one of these files?
I hope so :-)


> > Shall we simply merge as is for now, and add things when we need to?
> 
> I am fine with merging this PR.
Great!

PR #17, opened 2021-03-25
Add forcing submodule

stefsmeets:
Hi all, this PR adds the forcing submodule.

### DIFF
--- a/ewatercycle/forcing.py
+++ b/ewatercycle/forcing.py
@@ -1,0 +1,341 @@
+from esmvalcore.experimental import get_recipe
+from pathlib import Path
+
+FORCINGS = {
+    'ERA5': {
+        'dataset': 'ERA5',
+        'project': 'OBS6',
+        'tier': 3,
+        'type': 'reanaly',
+        'version': 1
+    },
+    'ERA-Interim': {
+        'dataset': 'ERA-Interim',
+        'project': 'OBS6',
+        'tier': 3,
+        'type': 'reanaly',
+        'version': 1
+    },
+}
+
+
+def update_hype(
+    data: dict,
+    *,
+    forcings: list = None,
+    startyear: int = None,
+    endyear: int = None,
+    shapefile: str = None,
+):
+    """
+    Update hype recipe data in-place.
+    
+    Parameters
+    ----------
+    forcings : list
+        List of forcings to use
+    startyear : int
+        Start year for the observation data
+    endyear : int
+        End year for the observation data
+    shapefile : str
+        Name of the shapefile to use.
+    """
+    preproc_names = ('preprocessor', 'temperature', 'water')
+
+    if shapefile is not None:
+        for preproc_name in preproc_names:
+            data['preprocessors'][preproc_name]['extract_shape'][
+                'shapefile'] = shapefile
+
+    if forcings is not None:
+        datasets = [FORCINGS[forcing] for forcing in forcings]
+        data['datasets'] = datasets
+
+    variables = data['diagnostics']['hype']['variables']
+    var_names = 'tas', 'tasmin', 'tasmax', 'pr'
+
+    if startyear is not None:
+        for var_name in var_names:
+            variables[var_name]['start_year'] = startyear
+
+    if endyear is not None:
+        for var_name in var_names:
+            variables[var_name]['end_year'] = endyear
+
+    return data
+
+
+def update_lisflood(
+    data: dict,
+    *,
+    forcings: list = None,
+    startyear: int = None,
+    endyear: int = None,
+    shapefile: str = None,
+    extract_region: dict = None,
+):
+    """
+    Update lisflood recipe data in-place.
+
+    Parameters
+    ----------
+    forcings : list
+        List of forcings to use
+    startyear : int
+        Start year for the observation data
+    endyear : int
+        End year for the observation data
+    basin : str
+        Name of the basin to use. Defines the shapefile.
+    extract_region : dict
+        Region specification, must contain `start_longitude`, 
+        `end_longitude`, `start_latitude`, `end_latitude`
+    """
+    preproc_names = ('general', 'daily_water', 'daily_temperature',
+                     'daily_radiation', 'daily_windspeed')
+
+    if shapefile is not None:
+        basin = Path(shapefile).stem
+        for preproc_name in preproc_names:
+            data['preprocessors'][preproc_name]['extract_shape'][
+                'shapefile'] = shapefile
+        data['diagnostics']['diagnostic_daily']['scripts']['script'][
+            'catchment'] = basin
+
+    if extract_region is not None:
+        for preproc_name in preproc_names:
+            data['preprocessors'][preproc_name][
+                'extract_region'] = extract_region
+
+    if forcings is not None:
+        datasets = [FORCINGS[forcing] for forcing in forcings]
+        data['datasets'] = datasets
+
+    variables = data['diagnostics']['diagnostic_daily']['variables']
+    var_names = 'pr', 'tas', 'tasmax', 'tasmin', 'tdps', 'uas', 'vas', 'rsds'
+
+    if startyear is not None:
+        for var_name in var_names:
+            variables[var_name]['start_year'] = startyear
+
+    if endyear is not None:
+        for var_name in var_names:
+            variables[var_name]['end_year'] = endyear
+
+    return data
+
+
+def update_marrmot(
+    data: dict,
+    *,
+    forcings: list = None,
+    startyear: int = None,
+    endyear: int = None,
+    shapefile: str = None,
+):
+    """
+    Update marmott recipe data in-place.
+
+    Parameters
+    ----------
+    forcings : list
+        List of forcings to use
+    startyear : int
+        Start year for the observation data
+    endyear : int
+        End year for the observation data
+    basin : str
+        Name of the basin to use. Defines the shapefile.
+    """
+
+    if shapefile is not None:
+        basin = Path(shapefile).stem
+        data['preprocessors']['daily']['extract_shape'][
+            'shapefile'] = shapefile
+        data['diagnostics']['diagnostic_daily']['scripts']['script'][
+            'basin'] = basin
+
+    if forcings is not None:
+        datasets = [FORCINGS[forcing] for forcing in forcings]
+        data['diagnostics']['diagnostic_daily'][
+            'additional_datasets'] = datasets
+
+    variables = data['diagnostics']['diagnostic_daily']['variables']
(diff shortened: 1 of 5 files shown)


ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
+from esmvalcore.experimental import get_recipe
+from pathlib import Path
+
+FORCINGS = {
```
Peter9192: I'd suggest changing this name to `DATASETS`. That more accurate, since the *actual* forcing data is the *output* of `generate_forcing`. So you could generate forcing data based on either the ERA5 or the ERA-Interim dataset (and possibly many others).

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+        'type': 'reanaly',
+        'version': 1
+    },
+}
+
+
+def update_hype(
+    data: dict,
+    *,
+    forcings: list = None,
```
Peter9192, line 25:
```suggestion
    datasets: list = None,
```
?

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+        'project': 'OBS6',
+        'tier': 3,
+        'type': 'reanaly',
+        'version': 1
+    },
+}
+
+
+def update_hype(
+    data: dict,
```
Peter9192: What is `data`?

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+):
+    """
+    Update hype recipe data in-place.
+    
+    Parameters
+    ----------
+    forcings : list
+        List of forcings to use
+    startyear : int
+        Start year for the observation data
```
Peter9192:
```suggestion
        Start year for the forcing data
```
It's not necessarily observation data. We could also force a hydrological model with output data from a climate model, for example.

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def update_lisflood(
+    data: dict,
+    *,
+    forcings: list = None,
+    startyear: int = None,
+    endyear: int = None,
+    shapefile: str = None,
+    extract_region: dict = None,
```
Peter9192, line 78: In the case of listflood (and maybe other models as well), we use both extract region *and* extract shape. I think the extract region settings in these cases is based on the extent of the shapefile plus some padding. This is what we used to do in the preprocessing notebooks in the comparison. 

I think it would make sense to do that padding also inside the forcing module. That would make the interface for all models more uniform and save hassle for the hydrologist, who shouldn't have to worry about using multiple preprocessors to eventually get his shapefile cut out. 

For example, from the listflood preprocessing notebook in the comparison: 

```python
def get_extents(shapefile, pad=0):
    """Get lat/lon extents from shapefile and add padding."""
    shape = fiona.open(shapefile)
    x0, y0, x1, y1 = [
        shply.geometry.shape(p["geometry"]).bounds for p in shape
    ][0]
    x0 = round((x0 - pad), 1)
    y0 = round((y0 - pad), 1)
    x1 = round((x1 + pad), 1)
    y1 = round((y1 + pad), 1)
    return x0, y0, x1, y1
```

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+    Parameters
+    ----------
+    forcings : list
+        List of forcings to use
+    startyear : int
+        Start year for the observation data
+    endyear : int
+        End year for the observation data
+    basin : str
+        Name of the basin to use. Defines the shapefile.
```
Peter9192, line 94: It would be nice to make this more uniform accross all models. Either pass the basin name, or the shapefile. Both have advantages or disadvantages, but I'm inclined towards the latter as it gives more flexibility for the user to add his own experiments.

BTW I think we agreed (once upon a time) that we prefer the term 'catchment' over 'basin'. See https://github.com/eWaterCycle/era5-comparison/blob/master/CONTRIBUTING.md#naming. This is not a criticism, but just a heads up if you ever encounter the term basin you might want to realize this is an ancient artefact of an old age....

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    if endyear is not None:
+        for var_name in var_names:
+            variables[var_name]['end_year'] = endyear
+
+    return data
+
+
+def update_marrmot(
+    data: dict,
```
Peter9192: Why pass only the data, and not just the recipe object itself?

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+    startyear: int = None,
+    endyear: int = None,
+    shapefile: str = None,
+):
+    """
+    Update marmott recipe data in-place.
+
+    Parameters
+    ----------
+    forcings : list
```
Peter9192: I noticed that the `data` argument is not explained. Maybe if we just pass the recipe object, it becomes more explainable/understandable?

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def update_wflow(
+    data: dict,
+    *,
+    forcings: list = None,
+    startyear: int = None,
+    endyear: int = None,
+    extract_region: dict = None,
+    dem_file: str = None,
```
Peter9192, line 264: In a far future we might want to make it possible to derive the DEM file directly from a shapefile. For now I think it's unavoidable that we have a deviating call signature here.

ewatercycle/forcing.py
```diff
@@ -0,0 +1,341 @@
...
+        'update_func': update_pcrglobwb,
+    },
+    'wflow': {
+        'recipe_name': 'hydrology/recipe_wflow.yml',
+        'update_func': update_wflow,
+    },
+}
+
+
+def generate(model: str, **kwargs):
```
Peter9192, line 336: I really like the way you streamlined this code!

tests/test_forcing.py
```diff
@@ -0,0 +1,1025 @@
...
+import pytest
+from ewatercycle.forcing import update_marrmot, update_lisflood, update_hype, update_wflow, update_pcrglobwb
+
+TEST_INPUT_MARRMOT = (
+    {},
+    {
+        'startyear': 1990,
+        'endyear': 2018,
+        'forcings': ['ERA-Interim', 'ERA5'],
+        'shapefile': 'Meuse/Meuse.shp',
```
Peter9192, line 13: Where do these shapefiles come from? They're not part of the package, are they?

tests/test_forcing.py
```diff
@@ -0,0 +1,1025 @@
+from esmvalcore.experimental import get_recipe
```
Peter9192, line 1: How do I run these tests? Pytest is not part of the package/environment I freshly installed

tests/test_forcing.py
```diff
@@ -0,0 +1,1025 @@
...
+    },
+)
+
+TEST_INPUT_LISFLOOD = (
+    {},
+    {
+        'startyear': 1990,
+        'endyear': 1990,
+        'forcings': ['ERA-Interim', 'ERA5'],
+        'shapefile': 'Lorentz_Basin_Shapefiles/Meuse/Meuse.shp',
```
Peter9192, line 35: Same, where does this path lead to?

tests/test_forcing.py
```diff
@@ -0,0 +1,1025 @@
...
+        expected_data = recipe.data
+        data = copy.deepcopy(expected_data)
+        update_func(data, **params)
+        diff = deepdiff.DeepDiff(recipe.data, data)
+
+        NEW_TEST_CASES.append(diff)
+
+    NEW_TEST_CASES = tuple(NEW_TEST_CASES)
+
+    breakpoint()
```
Peter9192:
```suggestion
```

Peter9192 (changes requested):
Hey @stefsmeets really very nice work on streamlining the code to generate forcing data!! I like it very much, especially in conjuntion with the example notebook [here](https://github.com/eWaterCycle/technicalPaperExampleNotebooks/blob/main/preprocessing.ipynb). 

I do have a couple of suggestions. Perhaps you could already address some of them, but I don't think it's necessary to make it perfect at this stage. It's also just to point out some relevant context that you might find useful for the future. I'm happy to clarify if some things are unclear :-)

Peter9192:
Oh and should we do something about the broken tests?

tests/test_forcing.py
stefsmeets: Pytest is a dependency in `setup.py`, meaning it will be installed if you run `python setup.py develop`

tests/test_forcing.py
stefsmeets: These are the default values in the recipes on `esmvaltool`. I use these values to 1. see if the original recipes can be reproduced, 2. track any changes to the recipes.

ewatercycle/forcing.py
stefsmeets: All function calls are slightly different for that matter, we will get back to this at a later point 👍

ewatercycle/forcing.py
stefsmeets: I'm fine with either, but I understood from Niels that this is in debate.

ewatercycle/forcing.py
stefsmeets: I agree, but that is outside of scope for this PR. We will get back to this when we agree on the design for the API.

ewatercycle/forcing.py
stefsmeets: I copied this from the example from @nielsdrost , who used `forcings`. I don't mind either way.

ewatercycle/forcing.py
stefsmeets: I changed the name to `recipe_dict` to make it more clear and added the docstring.

tests/test_forcing.py
stefsmeets: This is also the default value in the `esmvaltool` recipe.

tests/test_forcing.py
Peter9192: ah okay check

tests/test_forcing.py
Peter9192: Hmm that's strange, I did `pip install -e .[develop]` and that didn't work.

stefsmeets:
> Oh and should we do something about the broken tests?

I guess adding `esmvaltool` as a dependency causes a bit of an issue here.

That should probably be fixed in travis.

Peter9192:
Okay so we decided to merge this and tackle any issues we still foresee in follow-up PRs! Thanks for making this great start with our refactoring operation! ;-)

PR #20, opened 2021-04-01
Remove references to parametersetdb from docs configuration

Peter9192:
This PR changes the references from `ewatercycle_parametersetdb` to plain `ewatercycle`. 

So far I only modified the config and makefiles, not the actual references in e.g. index.rst. I think Sarah was working on this file in a separate PR. There are many more references throughout the repo, e.g. also in sonarcloud config.


Peter9192:
Whoops, branched off the wrong branch.

PR #22, opened 2021-04-01
Remove parametersetdb reference from docs config and makefiles

Peter9192:
This PR changes the references from ewatercycle_parametersetdb to plain ewatercycle.

So far I only modified the config and makefiles, not the actual references in e.g. index.rst. I think Sarah was working on this file in a separate PR. There are many more references throughout the repo, e.g. also in sonarcloud config.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
-SPHINXPROJ    = ewatercycle_parametersetdb
+SPHINXPROJ    = ewatercycle
 SOURCEDIR     = .
 BUILDDIR      = _build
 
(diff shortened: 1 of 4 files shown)


sverhoeven (approved):
I am able to generate html docs. There are some warnings about missing dask/esmvalcore, but those should be resolved elsewhere.

Besides docs/index.rst whic has not been updated yet, I don't see any remnants of parametersetdb.

sverhoeven:
I build the docs with
```shell
git checkout fix_rtd 
git pull
python3 -m venv env
. env/bin/activate
pip install -r requirements.txt -e .
cd docs
make html
xdg-open _build/html/index.html 
```

Peter9192:
Thanks @sverhoeven for checking and for the hints about the local build. I updated the docs/index.rst as well. You're right, I think that was the only remnant. The remaining references I found are those that actually belong to the original code that's obviously still there 😅 .

nielsdrost:
Thanks @Peter9192 !

PR #26, opened 2021-04-02
Drop requirements.txt and python setup.py test

### LINKED ISSUE #24: Drop requirements.txt in favor of .[dev]
The https://github.com/eWaterCycle/ewatercycle/blob/master/requirements.txt and https://github.com/eWaterCycle/ewatercycle/blob/8473707e441a9361380fec8a5f9b879fc3cda9a1/setup.py#L59 are both describing developement dependencies. It would be nice if there was just one place for the dev dependencies.

Can the content of requirements.txt be moved to dev extras in setup.py?

Please also document in README or CONTRIBUTING.rst how they should be installed:
```
pip install -e .[dev]
```

### LINKED ISSUE #25: Replace python setup.py test with pytest
To run tests the `python setup.py test ` command is deprecated. It should be replaced with pytest command.

sverhoeven:
Fixes #24 
Fixes #25

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install dependencies
       shell: bash -l {0}
       run: |
-        pip install -e .[dev]
+        pip3 install -e .[dev]
     - name: Test with pytest
       run: pytest
       shell: bash -l {0}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,5 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -e .[dev]
-        pip install -r requirements.txt
     - name: Test with pytest
       run: pytest
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install dependencies
       shell: bash -l {0}
       run: |
-        pip install -e .[dev]
+        pip3 install -e .[dev]
     - name: Tests with coverage
       run: pytest --cov --cov-report term --cov-report xml --junitxml=xunit-result.xml
       shell: bash -l {0}
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -20,7 +20,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -e .[dev]
-        pip install -r requirements.txt
     - name: Tests with coverage
       run: pytest --cov --cov-report term --cov-report xml --junitxml=xunit-result.xml
     - name: Correct coverage paths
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.rst
```diff
@@ -40,10 +40,11 @@ You want to make some kind of change to the code base
...
 2. (**important**) wait until some kind of consensus is reached about your idea being a good idea;
 3. if needed, fork the repository to your own Github profile and create your own feature branch off of the latest master commit. While working on your feature branch, make sure to stay up to date with the master branch by pulling in changes, possibly from the 'upstream' repository (follow the instructions `here <https://help.github.com/articles/configuring-a-remote-for-a-fork/>`_ and `here <https://help.github.com/articles/syncing-a-fork/>`_);
-4. make sure the existing tests still work by running ``python setup.py test``;
-5. add your own tests (if necessary);
-6. update or expand the documentation;
-7. `push <http://rogerdudler.github.io/git-guide/>`_ your feature branch to (your fork of) the ewatercycle repository on GitHub;
-8. create the pull request, e.g. following the instructions `here <https://help.github.com/articles/creating-a-pull-request/>`_.
+4. Install package in editiable mode and its dependencies with ``pip3 install -e .[dev]``;
```
SarahAlidoost:
```suggestion
4. install the package in editable mode and its dependencies with ``pip3 install -e .[dev]``;
```

.github/workflows/ci.yml
```diff
@@ -30,7 +30,6 @@ jobs:
       shell: bash -l {0}
       run: |
         pip install -e .[dev]
```
SarahAlidoost: Can we use pip3 instead of pip like the command explained in CONTRIBUTING.rst.

.github/workflows/sonar.yml
```diff
@@ -24,7 +24,6 @@ jobs:
       shell: bash -l {0}
       run: |
         pip install -e .[dev]
```
SarahAlidoost: Can we use pip3 instead of pip like the command explained in CONTRIBUTING.rst.

setup.py
```diff
@@ -41,22 +41,23 @@
...
+            'pytest',
+            'pytest-cov',
+            # Linters
+            'pycodestyle',
+            'prospector[with_pyroma]',
+            'yapf',
+            'isort',
+            # dependency for `pytest`
+            'pytest-runner',
+            # dependencies for `cd docs && make html`
```
SarahAlidoost: I could not find the command `cd docs && make html`  in either contributing or documentation. I could build the documentation using the command `python setup.py build_sphinx -Ea`.  I suggest removing the command from the comment here.

SarahAlidoost (approved):
@sverhoeven thank you. I was able to install and run tests locally with this PR. I added some minor suggestions here.

setup.py
sverhoeven: Added line to CONTRIBUTING.rst and simplified comment in setup.py.

SarahAlidoost:
@sverhoeven shall I merge this PR?

sverhoeven:
Yep lets merge.

Merging it will cause more conflicts in #32 PR, but as it already has conflict, some more should not matter that much

PR #27, opened 2021-04-02
Adds CITIATION.cff and .zenodo.json

sverhoeven:


.zenodo.json
```diff
@@ -0,0 +1,42 @@
...
+            "orcid": "0000-0002-5413-9038"
+        },
+        {
+            "affiliation": "Netherlands eScience Center",
+            "name": "Kalverla, Peter",
+            "orcid": "0000-0002-5025-7862"
+        },
+        {
+            "affiliation": "Netherlands eScience Center",
+            "name": "Alidoost, Sarah",
```
SarahAlidoost:
```suggestion
            "name": "Alidoost, Fakhereh",
```

CITATION.cff
```diff
@@ -0,0 +1,42 @@
...
+    orcid: "https://orcid.org/https://orcid.org/0000-0002-5413-9038"
+  -
+    affiliation: "Netherlands eScience Center"
+    family-names: Kalverla
+    given-names: Peter
+    orcid: "https://orcid.org/https://orcid.org/0000-0002-5025-7862"
+  -
+    affiliation: "Netherlands eScience Center"
+    family-names: Alidoost
+    given-names: Sarah
```
SarahAlidoost:
```suggestion
    given-names: Fakhereh
```

CITATION.cff
```diff
@@ -0,0 +1,42 @@
...
+    orcid: "https://orcid.org/https://orcid.org/0000-0001-8407-6472"
+cff-version: "1.1.0"
+keywords:
+  - hydrology
+  - gis
```
SarahAlidoost: As far as I know, we don't use many "gis" tools in this package. Can we replace it with "bmi" for now?

SarahAlidoost (changes requested):
@sverhoeven thank you. I suggested some minor edits. Also, these two files should include other contributors i.e. Barbara, Rolf, and Jerom before the release. Do you think it is better to add them here or make an issue that we don't forget?

CITATION.cff
sverhoeven: OK, hope people find it for the right abbreviation (basic modeling interface not body mass index)

sverhoeven:
> @sverhoeven thank you. I suggested some minor edits. Also, these two files should include other contributors i.e. Barbara, Rolf, and Jerom before the release. Do you think it is better to add them here or make an issue that we don't forget?

Better to add them here and now. Creating a release without them included should not happen.

CITATION.cff
sverhoeven: done

SarahAlidoost (approved):
@sverhoeven thank you for updating the list. I approved the PR. However, I am not sure if the list of contributors in citation.cff is still complete. Let's ask @nielsdrost opinion on this too.

sverhoeven:
Thanks for reviewing, I created #42

PR #23, opened 2021-04-02
Add esmvaltool to github action and setup.py

SarahAlidoost:
This PR:
- adds esmvaltool to ci.yml. sonar.yml and setup.py 
- adds deepdiff (one of the test requirements) to the requirements.txt
-  revise readme

PR #28, opened 2021-04-02
Update forcing generator

stefsmeets:
Hi all, this PR tunes the forcing generator functions to:

1. Return a `ForcingData` object, which at present is a very shallow wrapper around the esmvalcore `RecipeOutput`
2. Rename `forcings` -> `dataset` as an input parameter
3. Take a single dataset instead of a list


Peter9192:
Hey @stefsmeets can you try to merge master? I think the CI was fixed (in #23) since you started working on this. Would be nice to see some green tests

ewatercycle/forcing_data.py
```diff
@@ -0,0 +1,9 @@
+class ForcingData(object):
```
Peter9192: Maybe a dumb question, but why put this in a separate file? If I understand it correctly this leads to the following hierarchy (e.g. in the documentation)
```
ewatercycle 
- forcing
  - generate
- forcing_data
  - ForcingData
```

Wouldn't it make more sense to have something like:

```
ewatercycle
- forcing
  - ForcingData
  - generate
```
?

Peter9192 (changes requested):
Looks good! Perhaps you could merge master to see if the tests pass, and maybe also merge #31 so we can preview the documentation build? Other than that I just have one question about how to structure the module.

ewatercycle/forcing_data.py
stefsmeets: `forcing.py` is already quite big, and I also expect `ForcingData` will not turn out to be a small class, so I put it in a different file. Besides that, it also gives some nice separation of concerns.

Later on we can move both files in to a sub directory to keep them together, i.e. `./ewatercycle/forcing/`

Peter9192: Indeed, I like that last suggestion. Why not do that right away?

Peter9192: Notice that Stefan also made a `models` module in #35

stefsmeets: Restructured the code a bit in the latest commit 👍

Peter9192 (approved):

Peter9192:
Thanks Stef!

PR #31, opened 2021-04-02
Fix RTD build and badges

Peter9192:
This PR fixes a few things that broke since we moved the documentation to a new RTD page (ewatercycle.readthedocs.io). 

- Badge for RTD now refers to the correct page
- Package build badge refers to the correct package
- Broken RTD build is fixed (by mocking dependencies instead of installing them)

closes #30

PR #35, opened 2021-04-06
Added AbstractModel class

### LINKED ISSUE #33: Create model abstract class
Modelled after pymt and requirements of notebooks.

sverhoeven:
Fixes #33

Docs check fails, but should be solved in #31

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,8 +4,8 @@ ewatercycle
 
 A Python package for running hydrological models.
 
-.. image:: https://github.com/eWaterCycle/parametersetdb/actions/workflows/ci.yml/badge.svg
-    :target: https://github.com/eWaterCycle/parametersetdb/actions/workflows/ci.yml
+.. image:: https://github.com/eWaterCycle/ewatercycle/actions/workflows/ci.yml/badge.svg
+    :target: https://github.com/eWaterCycle/ewatercycle/actions/workflows/ci.yml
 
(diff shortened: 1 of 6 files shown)


Peter9192 (commented):
Hey @sverhoeven cool stuff! I was wondering, what about stuff like 
```
grid_id = model.get_var_grid(variable)
shape = model.get_grid_shape(grid_id)
lon = model.get_grid_x(grid_id)
lat = model.get_grid_y(grid_id)
```
I couldn't find what we decided about that in the backlog item

sverhoeven:
> Hey @sverhoeven cool stuff! I was wondering, what about stuff like
> 
> ```
> grid_id = model.get_var_grid(variable)
> shape = model.get_grid_shape(grid_id)
> lon = model.get_grid_x(grid_id)
> lat = model.get_grid_y(grid_id)
> ```
> 
> I couldn't find what we decided about that in the backlog item

I think those should not be public, instead the `get_as_xarray()` (#1460) or `run()` (#1459), they will internally use those bmi functions.

Peter9192:
> I think those should not be public, instead the `get_as_xarray()` (#1460) or `run()` (#1459), they will internally use those bmi functions.

Okay! That's also fine with me 👍

Peter9192 (approved):
I guess the real test will follow once we start implementing concrete model classes.

sverhoeven:
Thanks for reviewing

PR #36, opened 2021-04-07
Add get_value_as_xarray method to model API

Peter9192:
This PR adds a method to the model API to get the values of a variable as `xarray.DataArray`

(also sorted the imports)

### DIFF
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -1,8 +1,9 @@
-from abc import abstractmethod, ABCMeta
+from abc import ABCMeta, abstractmethod
 from os import PathLike
-from typing import Tuple, Iterable, Any, Optional
+from typing import Any, Iterable, Optional, Tuple
 
 import numpy as np
+import xarray as xr
 from basic_modeling_interface import Bmi
 
 
@@ -68,6 +69,17 @@ class AbstractModel:
 
         """
         self.bmi.set_value(name, value)
+
+    @abstractmethod
+    def get_value_as_xarray(self, name: str) -> xr.DataArray:
+        """Get a copy values of the given variable as xarray DataArray.
+
+        The xarray object also contains coordinate information and additional
+        attributes such as the units.
+
+        Args: name: Name of the variable
+
+        """
 
     @property
     @abstractmethod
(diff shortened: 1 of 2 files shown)


tests/models/test_abstract.py
```diff
@@ -120,8 +129,19 @@ def test_time_step(bmi, model: MockedModel):
...
+
+def test_get_value_as_xarray(model: MockedModel):
+    expected = np.array([[1.0, 2.0]])
+
+    dataarray = model.get_value_as_xarray("Temperature")
+
+    assert_array_equal(dataarray.values, expected)
+    assert dataarray.name == "Temperature"
+    assert dataarray.units == "degC"
+    assert isinstance(dataarray, xr.DataArray)
```
sverhoeven: Could use http://xarray.pydata.org/en/stable/generated/xarray.testing.assert_allclose.html#xarray-testing-assert-allclose instead of multiple asserts.

ewatercycle/models/abstract.py
```diff
@@ -69,6 +70,17 @@ def set_value(self, name: str, value: np.ndarray) -> None:
         """
         self.bmi.set_value(name, value)
 
+    @abstractmethod
```
sverhoeven, line 71: We could have a default implementation here, but due to different grids of models, best to leave abstract

sverhoeven (approved):
Looks good to me.
Tests pass, generated docs look good.

Small suggestion about xarray assert

tests/models/test_abstract.py
Peter9192: Is there a reason you chose `allcose` instead of `equal`? I noticed you used that for the numpy arrays as well.

sverhoeven: Any computation could lead to slightly different floats on different machines. To prevent tests from failing when run on another machine I always use allclose. I use allclose all the time so I dont have to think about what the computation is and how reproducible it is.

sverhoeven: In this case the test just compares mocked against expected. No computation happens which could change the floats. So here using equal is fine.

sverhoeven (approved):
Test looks much nicer now.

Peter9192:
Conflict resolved, merging this now. Thanks for reviewing @sverhoeven !

PR #39, opened 2021-04-07
Add Wflow implementation to the API

Peter9192:
This PR adds an implementation of the Wflow model using grpc4bmi.


ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+        da.data = np.reshape(model.get_value(variable), shape)
+        da.time.values = num2date(model.get_current_time(), time_units)
+        return da.where(da != -999)
+
+    @property
+    def parameters(self):
+        """List the variable names that are available for this model."""
+        return self.model.get_output_var_names()
```
sverhoeven: This property was modeled after pymt, there it is used to default values for the setup() method. So the usage here is not inline with pymt.

I think it should return something like
```python
{
 'RiverRunoff': '2, m^3/s'm',
}
```
Was thinking about 
* start/end/step time, but those are restricted by what is available in parameterset/forcing, so having a default does not make sense and you already have getters for them.
* paths (to files in parameterset, forcing_data) in the config file that need to reflect current session, but those are dont have defaults. However pymt docs says [*Check the parameters property to verify that the model inputs
have been updated.*](https://github.com/csdms/pymt/blame/master/docs/usage.rst#L153-L154), could make sense to set them after setup() method so they can be retrieved later.
* reinit flag or warmup period

Peter9192:
Okay, I have a first working implementation of the WFlow class. I also added an example notebook. Before I continue, I have a few questions:

- What do you think about the way I specified the arguments for Wflow.setup()?
- Any thoughts on how we should specify the configuration settings? 
- SurfaceRunoff isn't getting any values. Any clues why?

ewatercycle/models/wflow.py
Peter9192: Oh right, that reference is very useful, I hadn't looked at that yet!

Peter9192: So this can also be used to interactively set parameters during the model setup phase... However, for wflow the parameters (like RiverRunoff) are not simple values but entire spatial fields, more like initial conditions... Not sure how this translates to the PyMT philosophy.

nielsdrost:
"SurfaceRunoff isn't getting any values. Any clues why?"

@Peter9192 The name of this variable changed from SurfaceRunoff to RiverRunoff (or the other way around) at some point. 

Are you using the correct wflow version for your ini file?

Peter9192:
> 
> 
> "SurfaceRunoff isn't getting any values. Any clues why?"
> 
> @Peter9192 The name of this variable changed from SurfaceRunoff to RiverRunoff (or the other way around) at some point.
> 
> Are you using the correct wflow version for your ini file?

@nielsdrost you're a hero, that was it!

image

ewatercycle/models/wflow.py
Peter9192: I now made it such that it shows the config parameters that are set through the ini file. This only works after the `model.setup()` is run, but I don't think we can get any closer to PyMT given the constraints of our models.

Peter9192:
Note that we agreed that the notebook serves as a test for now.
I decided to stick with just passing the path to the forcing data rather than an object, as the alternative at this stage would make the notebooks needlessly more complex.

One thing that I'd still like to change is to match the output type of `setup` (PathLike) with the input type of `initialize` (str), but this requires changed the interface.

ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,157 @@
...
+
+        Returns:
+            Path to config file and config dir
+        """
+        self._setup_cfg_dir(cfg_dir=cfg_dir)
+        self._setup_cfg_file(cfg_file=cfg_file, **kwargs)
+        self._setup_forcing(forcing_data=forcing_data)
+        self._start_container()
+
+        return self.cfg_dir, self.cfg_file
```
sverhoeven: Can dir and file be swapped? See https://pymt.readthedocs.io/en/latest/usage.html#model-setup
```suggestion
        return self.cfg_file, self.cfg_dir
```

ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,157 @@
...
+
+# from ewatercycle import CFG
+
+# mock CFG until the CFG PR is merged.
+CFG = {
+    "container_engine": "docker",
+    "singularity_images.wflow": "ewatercycle-wflow-grpc4bmi.sif",
+    "docker_images.wflow": "ewatercycle/wflow-grpc4bmi:latest",
+    "scratch_dir": "./"
+}
```
sverhoeven: #45 has been merged. Can this be removed and master branch be merged into this branch?

ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,157 @@
...
+            with self.cfg_file.open("w") as filename:
+                self.cfg.write(filename)
+            print(
+                f"Copied forcing data to {self.cfg_dir} and updated {self.cfg_file} accordingly."
+            )
+
+    def _start_container(self):
+        if CFG["container_engine"] == "docker":
+            self.bmi = BmiClientDocker(
+                image=CFG["docker_images.wflow"],
```
sverhoeven: The CFG has slightly different keys. Can you adjust?
```suggestion
                image=CFG["wflow.docker_image"],
```

ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,157 @@
...
+                image_port=55555,
+                work_dir=str(self.cfg_dir),
+                timeout=10,
+            )
+        elif CFG["container_engine"] == "singularity":
+            image = CFG["singularity_images.wflow"]
```
sverhoeven:
```suggestion
            image = CFG["wflow.singularity_image"]
```

ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,157 @@
...
+                "longitude": self.bmi.get_grid_y(grid),
+                "latitude": self.bmi.get_grid_x(grid),
+                "time": num2date(self.bmi.get_current_time(), time_units)
+            },
+            dims=["latitude", "longitude"],
+            name=name,
+            attrs={"units": self.bmi.get_var_units(name)},
+        )
+
+        return da.where(da != -999)
```
sverhoeven, line 119: Can get_value be overwritten here to also replace -999 with na?

```suggestion
        return da.where(da != -999)

    def get_value(name: str) -> np.ndarray:
         da = super(name).get_value()
         return da.where(da != -999)
```

ewatercycle/models/wflow.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @property
+    def parameters(self):
+        """List the configurable parameters for this model."""
+        if hasattr(self, "cfg"):
+            for section in self.cfg.sections():
+                print(section)
+                for option in self.cfg.options(section):
+                    print(f"    {option}: {self.cfg.get(section, option)}")
```
sverhoeven: Prints does not make a great library experience. 
Pymt returns a list of tuple pairs. Can you return something like
```
(f'{section}.{option}', self.cfg.get(section, option))
```

sverhoeven (changes requested):
Notebook looks good to me.

Having forcing_data seems untested, but notebooks dont make use of it so that is OK.

Peter9192:
> Having forcing_data seems untested, but notebooks dont make use of it so that is OK.

We just decided to leave out the custom forcing option for now. I'll go ahead and remove it.

ewatercycle/models/wflow.py
Peter9192: Done 👍

ewatercycle/models/wflow.py
Peter9192: Good spot, I swapped them

ewatercycle/models/wflow.py
Peter9192: Yes, done

ewatercycle/models/wflow.py
Peter9192: Adjusted

ewatercycle/models/wflow.py
Peter9192: Check

ewatercycle/models/wflow.py
Peter9192: Hmm I don't really like that. `get_value` directly interfaces the model fiels, whereas `get_value_as_xarray` is a convenience function that helps the user by tranforming it to something more user friendly.

I changed the inline comments, hope this is clearer. The ideal solution of course is to fix it in wflow.

Peter9192:
Thanks for the review @sverhoeven! I think I covered all of it. Could you do another pass?

ewatercycle/config/ewatercycle.yaml
```diff
@@ -1,7 +1,7 @@
...
 grdc_location: null
-container_engine: null
-singularity_dir: null
-output_dir: null
-wflow.singularity_image: null
-wflow.docker_image: null
+container_engine: docker
+singularity_dir: './'
+output_dir: './'
+wflow.singularity_image: "./ewatercycle-wflow-grpc4bmi.sif"
```
sverhoeven, line 6: Not sure if a file is a good default, could lead to unexpected errors. We could use the Docker Hub as well for singularity.
```suggestion
wflow.singularity_image: "docker://ewatercycle/wflow-grpc4bmi:latest"
```

ewatercycle/models/wflow.py
sverhoeven: Yep, comment in notebook clarifies it is the fault of the model not the ewatercycle or grpc4bmi package.

sverhoeven (approved):
Looks OK, except ewatercycle/config/ewatercycle.yaml

I think it should be conscious choice by the system admin or user to pick a container engine and images. Having these defaults will give errors like Docker is not installed or the latest wflow does not work for me.
Can you rollback so default file has all null's again?

PR #40, opened 2021-04-07
Add lisflood class

SarahAlidoost:


ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+from grpc4bmi.bmi_client_singularity import BmiClientSingularity
+
+from ewatercycle import CFG
+from ewatercycle.forcing.forcing_data import ForcingData
+from ewatercycle.models.abstract import AbstractModel
+from ewatercycle.parametersetdb.config import AbstractConfig
+
+
+@dataclass
+class LisfloodParameterSet:
```
Peter9192, line 24: Nice, I like this solution

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+    """A NetCDF file with model boundaries"""
+    config_template: PathLike
+    """Config file used as template for a lisflood run"""
+    lisvap_config_template: Optional[PathLike] = None
+    """Config file used as template for a lisvap run"""
+
+
+# LISFLOOD settings file reference maps prefixes
+# MapsName: {prefix_name, prefix}
+MAPS_PREFIXES = {
```
Peter9192, line 51: Not sure I understand what this is needed for.

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+    """
+
+    # unable to subclass with more specialized arguments so ignore type
+    def setup(self,  # type: ignore
+              forcing: Union[ForcingData, PathLike],
+              parameterset: LisfloodParameterSet,
+              work_dir: PathLike = None) -> Tuple[PathLike, PathLike]:
+        """Configure model run
+
+        If forcing is missing evaporation files (e0, es0, et0) then the :py:meth:`run_lisvap` function should be run.
```
Peter9192: This instruction is a bit unclear. If my understanding is correct, then perhaps you could rephrase as:

```suggestion
        If evaporation files (e0, es0, et0) are not included in the forcing, then :py:meth:`run_lisvap` function should be called before setup.
```

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+        bmi (Bmi): Basic Modeling Interface object
+        work_dir (PathLike): Working directory for the model where it can read/write files
+
+    Example:
+        See examples/lisflood.ipynb in `ewatercycle repository <https://github.com/eWaterCycle/ewatercycle>`_
+    """
+
+    # unable to subclass with more specialized arguments so ignore type
+    def setup(self,  # type: ignore
+              forcing: Union[ForcingData, PathLike],
```
Peter9192, line 71: Can forcing be part of the parameterset? Or should it always be passed seperately?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+    'ET0Maps': {'name': 'PrefixET0', 'value': 'et0'},
+}
+
+
+class Lisflood(AbstractModel):
+    """eWaterCycle implementation of Lisflood hydrological model.
+
+    Attributes:
+        bmi (Bmi): Basic Modeling Interface object
+        work_dir (PathLike): Working directory for the model where it can read/write files
```
Peter9192, line 63: Why list (only) these and not e.g. `self.parameterset`?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+            )
+        else:
+            raise ValueError(
+                f"Unknown container technology in CFG: {CFG['container_engine']}"
+            )
+        return Path(config_file), self.work_dir
+
+    def _check_work_dir(self, work_dir):
+        """Check and optionally create the work directory."""
+        self.work_dir = _generate_workdir(work_dir)
```
Peter9192: Why is this a separate function? Isn't it a bit overkill to call `_check_work_dir` if the only thing it does is call `_generate_workdir`?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def _check_forcing(self, forcing):
+        """"Check forcing argument."""
```
Peter9192: Can we make this docstring slightly more informative?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+            self.forcing_files = dict()
+            for forcing_file in self.forcing_dir.glob('*.nc'):
+                dataset = xr.open_dataset(forcing_file)
+                # TODO check dataset was created by ESMValTool, to make sure var names are as expected
+                var_name = list(dataset.data_vars.keys())[0]
+                self.forcing_files[var_name] = forcing_file.name
+                # get start and end date of time dimension
+                # TODO converting numpy.datetime64 to datetime object is ugly, find better way
+                self.start = datetime.utcfromtimestamp(dataset.coords['time'][0].values.astype('O') / 1e9)
+                self.end = datetime.utcfromtimestamp(dataset.coords['time'][-1].values.astype('O') / 1e9)
```
Peter9192, line 141: Maybe something to think about: do we want the forcing data to determine the start/end time, or do we want the used to set the start/end time independently of the forcing?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+    if work_dir is None:
+        scratch_dir = CFG['output_dir']
+        # TODO this timestamp isnot safe for parallel processing
+        timestamp = time.strftime("%Y%m%d_%H%M%S")
+        work_dir = Path(scratch_dir) / f'lisflood_{timestamp}'
+        work_dir.mkdir(parents=True, exist_ok=True)
+    return work_dir
+
+
+class XmlConfig(AbstractConfig):
```
Peter9192, line 387: Should we move this to `ewatercycle/parametersetdb/config` so it can be reused by other (future) model classes as well?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+                if prefix['name'] in textvar_name:
+                    textvar.set(
+                        "value",
+                        f"lisflood_{prefix['value']}_{timestamp}",
+                    )
+                if map_var in textvar_name:
+                    textvar.set('value', f"$(PathOut)/$({prefix['name']})")
+
+        # Write to new setting file
+        lisflood_file = f"{self.work_dir}/lisflood_setting_{timestamp}.xml"
```
Peter9192: Why do we need a timestamp in the settings filename?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+        lisvap_file = f"{self.work_dir}/lisvap_setting_{timestamp}.xml"
+        cfg.save(lisvap_file)
+        return lisvap_file
+
+    # TODO take this out of the class?
+    def run_lisvap(self, forcing: PathLike) -> Tuple[int, bytes, bytes]:
+        """Run lisvap to generate evaporation input files
+
+        Args:
+            forcing: Path to re-indexed forcings
```
Peter9192: What do you mean by "re-indexed"?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+                    )
+                if map_var in textvar_name:
+                    textvar.set('value', f"$(PathOut)/$({prefix['name']})")
+
+        # Write to new setting file
+        lisflood_file = f"{self.work_dir}/lisflood_setting_{timestamp}.xml"
+        cfg.save(lisflood_file)
+        return lisflood_file
+
+    def _create_lisvap_config(self) -> str:
```
Peter9192, line 205: There seems to be quite a bit of duplication here. Is there a nicer way to use shared config settings between lisflood and lisvap? Or are they too different?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+            },
+            dims=["latitude", "longitude"],
+            name=name,
+            attrs={"units": self.bmi.get_var_units(name)},
+        )
+
+        return da
+
+    @property
+    def parameters(self) -> Iterable[Tuple[str, Any]]:
```
Peter9192, line 333: I don't really understand why these would be the 'configurable' parameters. Wouldn't it make more sense if it listed the options in the lisflood config file?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+        """List the parameters for this model."""
+        if self.forcing_dir and self.work_dir:
+            return {
+                'forcing_dir': self.forcing_dir,
+                'work_dir': self.work_dir,
```
Peter9192, line 339: I have a bit of an internal conflict about the naming of `work_dir` (which is what grpc4bmi uses) and `cfg_dir` (which is what PyMT uses). I'm inclined to go with the latter for everything that the user interacts with, to stay in line with PyMT.

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+            }.items()
+        return []
+
+
+def reindex_forcings(mask_map: PathLike, forcing: ForcingData, output_dir: PathLike = None) -> PathLike:
```
Peter9192, line 344: I'm not sure I understand this. If this is necessary, doesn't that mean that the forcing is just not 'ready' yet as it comes out of ESMValTool?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def _check_work_dir(self, work_dir):
+        """Check and optionally create the work directory."""
+        self.work_dir = _generate_workdir(work_dir)
+
+    def _check_forcing(self, forcing):
+        """"Check forcing argument."""
+
+        # TODO check if mask has same grid as forcing files,
+        # if not warn users to run reindex_forcings
```
Peter9192, line 129: Should the user be concerned with this or should it be done automatically?

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,398 @@
...
+        encoding = {var_name: {"zlib": True, "complevel": 4, "chunksizes": (1,) + dataset[var_name].shape[1:]}}
+        dataset.reindex(
+                    {"lat": mask["lat"], "lon": mask["lon"]},
+                    method="nearest",
+                    tolerance=1e-2,
+                ).to_netcdf(out_fn, encoding=encoding)
+    return output_dir
+
+
+def _generate_workdir(work_dir: PathLike = None) -> PathLike:
```
Peter9192, line 371: Why is this a function rather than a method?

setup.py
```diff
@@ -21,6 +21,7 @@
     install_requires=[
         'basic_modeling_interface',
         'esmvaltool',
+        'grpc4bmi>=0.2.12,<0.3',
```
Peter9192, line 24: I think this is already part of master, no?

tests/models/test_lisflood.py
```diff
@@ -0,0 +1,111 @@
...
+from esmvalcore.experimental.recipe_output import DataFile
+from grpc4bmi.bmi_client_singularity import BmiClientSingularity
+
+from ewatercycle import CFG
+from ewatercycle.forcing.forcing_data import ForcingData
+from ewatercycle.parametersetdb.datafiles import SubversionCopier
+from ewatercycle.models.lisflood import Lisflood, LisfloodParameterSet
+
+
+@pytest.fixture
```
Peter9192, line 17: For my understanding, could you explain what we are really testing here? Since we're mostly testing against mocks and fixtures..

Peter9192 (changes requested):
Hey @SarahAlidoost really impressive work here! The lisflood class seems to be quite a bit more involved than wflow! 

Please see my other comments below and let me know if anything is unclear.

Peter9192:
I left some comments via reviewnb as well, but I don't see them show up here. Please see: https://app.reviewnb.com/eWaterCycle/ewatercycle/blob/add_lisflood_class/examples%2Flisflood.ipynb%2F/discussion

ewatercycle/models/lisflood.py
SarahAlidoost: An explanation is added.

ewatercycle/models/lisflood.py
SarahAlidoost: It can be, I added it as a TODO here.

ewatercycle/models/lisflood.py
SarahAlidoost: I think because `parameterset` is given by users to the `setup()`. @sverhoeven what do you think?

ewatercycle/models/lisflood.py
SarahAlidoost: The `_generate_workdir` is also used in `reindex_forcings ` that is not part of the model class.

ewatercycle/models/lisflood.py
SarahAlidoost: Sure, it is revised.

ewatercycle/models/lisflood.py
SarahAlidoost: Good question. If the forcing is the only temporal data given to lisflood, it makes sense that the start/end time of forcing is used in configuration and thus in initialization. Users can choose any start/end time (in forcing time range) while runs the model.

ewatercycle/models/lisflood.py
SarahAlidoost: We (@sverhoeven and I) discussed this. `source` is a path in this class whereas it is an URL in other classes in `ewatercycle/parametersetdb/config`.

ewatercycle/models/lisflood.py
SarahAlidoost: Good point, it is removed.

ewatercycle/models/lisflood.py
SarahAlidoost: it is removed.

ewatercycle/models/lisflood.py
SarahAlidoost: I don't think so. lisflood and lisfvap are two separate modules. I think that it is more clear if they have two separate config files, for more info on lisflood documentation, see https://ec-jrc.github.io/lisflood-model/.

ewatercycle/models/lisflood.py
SarahAlidoost: I agree, @sverhoeven what do you think?

ewatercycle/models/lisflood.py
SarahAlidoost: I agree. In [PyMT usage](https://pymt.readthedocs.io/en/latest/usage.html), the variable is named as `cfg_dir`. However, it is referred to **run directory** in the text. Let's make an issue for this and discuss it in a design session.

ewatercycle/models/lisflood.py
SarahAlidoost: This is the last part in the lisflood preprocessing notebook in comparison. The grid of forcing data generated by ESMValTool does not match the lisflood grid. The lisflood grid is given by a `model_mask` file that is available in its parameterset.

ewatercycle/models/lisflood.py
SarahAlidoost: I think users.

ewatercycle/models/lisflood.py
SarahAlidoost: The `_generate_workdir` is also used in `reindex_forcings` that is not part of the model class.

setup.py
SarahAlidoost: Good catch, yes, it is sorted now.

tests/models/test_lisflood.py
SarahAlidoost: @sverhoeven could you please reply to this question? thanks.

SarahAlidoost:
> Hey @SarahAlidoost really impressive work here! The lisflood class seems to be quite a bit more involved than wflow!
> 
> Please see my other comments below and let me know if anything is unclear.

@Peter9192 thank you for reviewing and helpful comments. For some of your questions on code and reviewnb, I asked @sverhoeven to help with answers. I have to re-run the notebook and push the results. I will inform you when it is done.

ewatercycle/models/lisflood.py
Peter9192: Oh right, yes, I stumbled upon that as well. Perhaps we can try to make that more flexible in `parametersetdb`?

ewatercycle/models/lisflood.py
sverhoeven: Correct, anything given by user can be treated as private vars inside class, while things the class creates that are useful for user are listed here.

ewatercycle/models/lisflood.py
sverhoeven: Yep, that will make the forcing argument optional.

ewatercycle/models/lisflood.py
sverhoeven: The forcing gives a maximum range. A user can use update() to stop early, but now can not start later. Supplying a start date to setup could help, but not in scope for current use cases.

ewatercycle/models/lisflood.py
sverhoeven: Now setup and parameters are sides of the same coin. So any change we make here should also be done in setup().
The forcing dir and working dir become part of the config file.

Which arguments should the setup have?  In wflow it is very dynamic but luckily 2 levels deep. In Lisflood there is a more complicated config with differently shaped sections and nestings.

tests/models/test_lisflood.py
sverhoeven: The only thing not real is the bmi object to make tests quick. This test exercises setup() method, but does not check content of config file.

As we switched from unit testing against to lisflood use case (https://github.com/ec-jrc/lisflood-usecases/tree/master/LF_lat_lon_UseCase) to using the notebook for testing. We stopped further developments here.

ewatercycle/models/lisflood.py
SarahAlidoost: We added a TODO item in setup().

ewatercycle/models/lisflood.py
sverhoeven: The LisFlood model is the only model that uses XML, when we add a another model that uses XML then we can move it over

ewatercycle/models/lisflood.py
```diff
@@ -0,0 +1,400 @@
...
+
+    Args:
+        mask_map: Path to NetCDF file used a boolean map that defines model boundaries.
+        forcing: Forcing data from ESMValTool
+        output_dir: Directory where to write the re-indexed files, given by user or created for user
+
+    Returns:
+        Output dir with re-indexed files.
+    """
+    output_dir = _generate_workdir(output_dir)
```
Peter9192, line 355: I don't understand. Here you pass `output_dir` as the value for the `work_dir` argument. But if I read `_generate_workdir` correctly, it will then just return output_dir, right? So it will not do anything.

ewatercycle/models/lisflood.py
Peter9192: See my new comment on this usage from withing `reindex_forcings`. It seems it's not doing anything. I still don't understand.

ewatercycle/models/lisflood.py
Peter9192: Created an issue: https://github.com/eWaterCycle/ewatercycle/issues/58

ewatercycle/models/lisflood.py
Peter9192: I'm not sure either. I just think the philosophy of PyMT is that these parameters are the things you'd like to change about the model configuration. So more like the default values from the [tables here](https://ec-jrc.github.io/lisflood-model/3_01_optLISFLOOD_overview/). But I agree that there is probably no neat way to do that right now. Perhaps one way would be to do it through the ParameterSet class you created. 

I think I'd prefer a NotImplementedError over the current solution, because I'm not very happy about e.g. having to pass a work_dir on setup and this feels like advertising that. 

Also note that it's currently not the same as the arguments to `model.setup`: parameterset is missing.

SarahAlidoost:
> Hey @SarahAlidoost really impressive work here! The lisflood class seems to be quite a bit more involved than wflow!
> 
> Please see my other comments below and let me know if anything is unclear.

Hey @Peter9192, we (@sverhoeven and I) addressed your comments. Please have another look, thanks.

ewatercycle/models/lisflood.py
SarahAlidoost: right, the function `_check_work_dir` is removed now.

ewatercycle/models/lisflood.py
SarahAlidoost: right, the function `_check_work_dir` is now removed.

Peter9192 (approved):

Peter9192:
Thansk @SarahAlidoost and @sverhoeven , great work!

PR #41, opened 2021-04-08
Make AbstractModel un-constuctable

### LINKED ISSUE #37: Possible to instantiate abstractclass while it shouldn't?
I think there might be a bug in the definition of the model API, as I seem to be able to instantiate classes for which not all methods are defined. Maybe I just don't understand it well enough though. 

Current implementation MWE:
```python
class AbstractModel:
    __metaclass__ = ABCMeta   # this is how it's currently implemented
    @abstractmethod
    def setup(self): pass

AbstractModel()   # works (while it shouldn't)
```

I think it can be fixed like this:
```python
class AbstractModel(metaclass=ABCMeta):
    @abstractmethod
    def setup(self): pass

AbstractModel()  # gives TypeError (as it should)
```

Was there a reason to do it like the first MWE?

sverhoeven:
Fixes #37

### DIFF
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -6,14 +6,12 @@ import numpy as np
 from basic_modeling_interface import Bmi
 
 
-class AbstractModel:
+class AbstractModel(metaclass=ABCMeta):
     """Abstract class of a eWaterCycle model.
 
     Attributes
         bmi (Bmi): Basic Modeling Interface object
     """
-    __metaclass__ = ABCMeta
-
     def __init__(self):
         self.bmi: Optional[Bmi] = None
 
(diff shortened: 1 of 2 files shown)


Peter9192 (commented):
Thanks for looking into this @sverhoeven !

This now fails, as expected: 
```
AbstractModel()
# TypeError: Can't instantiate abstract class AbstractModel with abstract methods parameters, setup
```

Similarly
```
class Wflow(AbstractModel):
   def foo(self): pass

Wflow()
# TypeError: Can't instantiate abstract class Wflow with abstract methods parameters, setup
```

I just wonder if I get this correct, since I didn't want the Wflow class to be abstract as well.

sverhoeven:
Warning makes sense, every subclass should have a setup() and parameters property implemented if you dont want it to be abstract.

Peter9192 (approved):

Peter9192:
Okay great! Thanks Stefan!

PR #45, opened 2021-04-08
Add module for reading configuration file

stefsmeets:
Hi all, this PR adds a module for reading a configuration file.

The default values are stored in `ewatercycle.yaml` in the root directory, and then it updates the defaults with the user config from `~/ewatercycle/ewatercycle.yaml` or `/etc/ewatercycle.yaml`. It searches in that order, and loads the file it finds first. The user config is not required and does not have to be complete.

### Todo

- [x] Write tests


stefsmeets:
@Peter9192 Documentation is failing with `ModuleNotFoundError: No module named 'ruamel'`. I cannot reproduce this error locally, and I already tried a few things. Since you did some debugging with `pydoc_auto_imports`, do you have any idea where to look?

stefsmeets:
Hi @sverhoeven @Peter9192 , this PR is ready for review, can one of you to have a look? 👍

ewatercycle/config/_config_object.py
```diff
@@ -0,0 +1,109 @@
...
+        user_config = source / filename
+        if user_config.exists():
+            return user_config
+    return None
+
+
+FILENAME = 'ewatercycle.yaml'
+
+SOURCES = (
+    Path.home() / '.ewatercycle',
```
sverhoeven: Can we change to  `~/.config/ewatercycle` this where the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) specifies the users config files should be ?
```suggestion
    Path.home() / '.config' / '.ewatercycle',
```

And if XDG_CONFIG_HOME env var could be used that would be even nicer.

ewatercycle/config/_config_object.py
```diff
@@ -0,0 +1,109 @@
...
+    Path('/etc'),
+)
+
+USER_CONFIG = find_user_config(SOURCES, FILENAME)
+DEFAULT_CONFIG = Path(ewatercycle.__file__).parents[1] / FILENAME
```
sverhoeven: When package is installed in a fresh virtualenv with `pip install --no-deps git+https://github.com/eWaterCycle/ewatercycle.git@config` the `<repo root>/ewatercycle.yaml` is not included. So this will fail when package is installed from pypi.

The default config should be moved to ewatercycle/config/ewatercycle.yaml and the setup.py must be adjusted to included into the distros. See https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html

ewatercycle/config/_validators.py
```diff
@@ -0,0 +1,141 @@
...
+
+_validators = {
+    'esmvaltool_config': validate_path_or_none,
+    'grdc_location': validate_path_or_none,
+    'container_engine': validate_string_or_none,
+    'singularity_dir': validate_path_or_none,
+    'output_dir': validate_path_or_none,
+    'ewatercycle_config': validate_path_or_none,
+    # wflow specific
+    'wflow.singularity_images': validate_string,
```
sverhoeven: Why is it plural (images)? As a system administrator I want to give my users a single image they can use per model. Can `singularity_images` and `docker_images` be made singular, `singularity_image` and `docker_image`.

ewatercycle/config/_validators.py
```diff
@@ -0,0 +1,141 @@
...
+
+validate_path_or_none = _make_type_validator(validate_path, allow_none=True)
+
+validate_pathlist = _listify_validator(validate_path,
+                                       docstring='Return a list of paths.')
+
+_validators = {
+    'esmvaltool_config': validate_path_or_none,
+    'grdc_location': validate_path_or_none,
+    'container_engine': validate_string_or_none,
```
sverhoeven, line 134: As container_engine determines which engine to use for every model we do not need to have a separate image for Docker and Singularity. We could simply have `wflow.image` instead of `wflow.singularity_images` and `wflow.docker_images`.

ewatercycle/config/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""config module.
+
+.. data:: CFG
+
+    eWaterCycle configuration object.
```
sverhoeven: Can we document where CFG gets its content from? Something like
```suggestion
    eWaterCycle configuration object.
    
    The configuration is loaded from 
    1. ~/.ewatercycle/ewatercycle.yaml
    2. /etc/ewatercycle/ewatercycle.yaml
    3. Fall back to empty configuration
    
    The `ewatercycle.yaml` is formatted in YAML and could for example look like
    
    .. code-block:: yaml
    
      esmvaltool_config: ~/.esmvaltool/config-user.yml
      grdc_location: /data/grdc
      container_engine: singularity
      singularity_dir: /data/singularity-images
      output_dir: /scratch
      # Created with cd  /data/singularity-images && singularity pull docker://ewatercycle/wflow-grpc4bmi:2020.1.1
      wflow.singularity_images: wflow-grpc4bmi_2020.1.1.sif
      wflow.docker_images: ewatercycle/wflow-grpc4bmi:2020.1.1
```

ewatercycle/config/_validated_config.py
```diff
@@ -0,0 +1,91 @@
...
+
+class MissingConfigParameter(UserWarning):
+    """Config parameter is missing."""
+
+
+# The code for this class was take from matplotlib (v3.3) and modified to
+# fit the needs of eWaterCycle. Matplotlib is licenced under the terms of
+# the the 'Python Software Foundation License'
+# (https://www.python.org/psf/license)
+class ValidatedConfig(MutableMapping):
```
sverhoeven, line 24: I tried to feed my own config and got unexpected `InvalidConfigParameter: `container_engine` is not a valid config parameter.` exception

mine.yml
```yaml
esmvaltool_config: ~/.esmvaltool/config-user.yml
grdc_location: /data/grdc
container_engine: singularity
singularity_dir: /data/singularity-images
output_dir: /scratch
# Created with cd  /data/singularity-images && singularity pull docker://ewatercycle/wflow-grpc4bmi:2020.1.1
wflow.singularity_images: wflow-grpc4bmi_2020.1.1.sif
wflow.docker_images: ewatercycle/wflow-grpc4bmi:2020.1.1
```

```python
In [1]: from ewatercycle.config import Config
In [2]: c = Config._load_user_config('mine.yml')
In [3]: c
Out[3]: 
Config({'container_engine': 'singularity',
        'esmvaltool_config': PosixPath('/home/verhoes/.esmvaltool/config-user.yml'),
        'ewatercycle_config': PosixPath('/home/verhoes/git/eWaterCycle/ewatercycle/mine.yml'),
        'grdc_location': PosixPath('/data/grdc'),
        'output_dir': PosixPath('/scratch'),
        'singularity_dir': PosixPath('/data/singularity-images'),
        'wflow.docker_images': 'ewatercycle/wflow-grpc4bmi:2020.1.1',
        'wflow.singularity_images': 'wflow-grpc4bmi_2020.1.1.sif'})

In [5]: from ewatercycle.config._validated_config import ValidatedConfig

In [6]: ValidatedConfig(c)
---------------------------------------------------------------------------
InvalidConfigParameter                    Traceback (most recent call last)
<ipython-input-6-d3b3454660ac> in <module>
----> 1 ValidatedConfig(Config._load_user_config('mine.yml'))

~/git/eWaterCycle/ewatercycle/ewatercycle/config/_validated_config.py in __init__(self, *args, **kwargs)
     31         super().__init__()
     32         self._mapping = {}
---> 33         self.update(*args, **kwargs)
     34 
     35     def __setitem__(self, key, val):

~/miniconda38/envs/ewatercycle/lib/python3.9/_collections_abc.py in update(self, other, **kwds)
    938         if isinstance(other, Mapping):
    939             for key in other:
--> 940                 self[key] = other[key]
    941         elif hasattr(other, "keys"):
    942             for key in other.keys():

~/git/eWaterCycle/ewatercycle/ewatercycle/config/_validated_config.py in __setitem__(self, key, val)
     40             raise InvalidConfigParameter(f"Key `{key}`: {verr}") from None
     41         except KeyError:
---> 42             raise InvalidConfigParameter(
     43                 f"`{key}` is not a valid config parameter.") from None
     44 

InvalidConfigParameter: `container_engine` is not a valid config parameter.

```

Am I using it wrong?

sverhoeven (changes requested):
Kind of works, but needs a bit more work.
See inline comments for questions and suggestions.

ewatercycle/config/_validated_config.py
sverhoeven: I expected errors from the non existing paths.

ewatercycle/config/_config_object.py
stefsmeets: Cool, I was not aware of this specification. As in `Path.home() / XDG_CONFIG_HOME / '.ewatercycle',` ?

On my instance `XDG_CONFIG_HOME` is not defined, so it would fall back to `~/.ewatercycle/`, instead of `~/.config/` though.

ewatercycle/config/_config_object.py
stefsmeets: Thanks for pointing this out!
I like to put the config in the root directory to make it easy to find. It's also useful to have it near the code, so I have moved it to the `config` submodule and updated `setup.py` 👍

ewatercycle/config/_validators.py
stefsmeets: Yeah, that makes more sense. I think I will leave it to @Peter9192 to update these values as necessary. I copied them from his PR to see if this could be a nice way to retrieve nested variables via `subset = CFG.get_subset('wflow')`.

ewatercycle/config/_validators.py
stefsmeets: Good point. Let's see what comes out of the model development, and then we can update the config accordingly. I think this is better addressed there.

ewatercycle/config/_validated_config.py
stefsmeets: You are using it correctly, in the sense that this works as expected:

```python
In [1]: from ewatercycle.config import Config
In [2]: c = Config._load_user_config('mine.yml')
In [3]: c
Out[3]: 
Config({'container_engine': 'singularity',
        'esmvaltool_config': PosixPath('/home/verhoes/.esmvaltool/config-user.yml'),
        'ewatercycle_config': PosixPath('/home/verhoes/git/eWaterCycle/ewatercycle/mine.yml'),
        'grdc_location': PosixPath('/data/grdc'),
        'output_dir': PosixPath('/scratch'),
        'singularity_dir': PosixPath('/data/singularity-images'),
        'wflow.docker_images': 'ewatercycle/wflow-grpc4bmi:2020.1.1',
        'wflow.singularity_images': 'wflow-grpc4bmi_2020.1.1.sif'})
```

Note how the paths are loaded as `pathlib.PosixPath` objects. So it validates the type (i.e., does your string look like something that could be a path), but it does not check if that paths exist. We discussed this briefly, and we agreed that this is not necessary. If we need the validator to check whether the paths exist after all and raise a warning/exception, I can add it.

```python
In [5]: from ewatercycle.config._validated_config import ValidatedConfig
In [6]: ValidatedConfig(c)
```
This will not work, because `ValidatedConfig` is the base class that `Config` inherits from. It does not know which config parameters are associated with the config. `Config` has an attribute `._validate`, which is a dict of validators to apply to every key/value pair. This one is empty in `ValidatedConfig`.

stefsmeets:
Hi @sverhoeven I updated the PR, could you have another look to see if I addressed your comments?

setup.py
```diff
@@ -30,6 +30,9 @@
         # 'grpc4bmi>=0.2.12,<0.3',
     ],
     packages=find_packages(),
+    package_data={
+        "": ["*.yaml"],
+    },
     include_package_data=True,
```
sverhoeven: Yaml file is included in bdist_wheel, but missing in sdist.  To also have yaml file in sdist remove include_package_data field.
```suggestion
```

ewatercycle/config/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    InvalidConfigParameter: `output_directory` is not a valid config parameter.
+
+Or, if the value entered cannot be converted to the expected type:
+
+.. code-block:: python
+
+    >>> CFG['output_dir'] = 123
+    InvalidConfigParameter: Key `output_dir`: Expected a path, but got 123
+
+By default, the config is loaded from the default location (i.e. ``~/.config/.ewatercycle/config-user.yml``).
```
sverhoeven: Config path not the same as code. Code looks for ~/.config/.ewatercycle/ewatercycle.yaml . Please make consistent.
```suggestion
By default, the config is loaded from the default location (i.e. ``~/.config/.ewatercycle/ewatercycle.yaml``).
```

ewatercycle/config/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    >>> CFG.reload()
+
+.. data:: CFG
+
+eWaterCycle configuration object.
+
+The configuration is loaded from:
+
+ 1. ``~/.ewatercycle/$XDG_CONFIG_HOME/.ewatercycle/ewatercycle.yaml``
+ 2. ``~/.ewatercycle/.config/.ewatercycle/ewatercycle.yaml``
```
sverhoeven: Config path not the same as code. Code looks for ~/.config/.ewatercycle/ewatercycle.yaml . Please make consistent.
```suggestion
 1. ``~/$XDG_CONFIG_HOME/.ewatercycle/ewatercycle.yaml``
 2. ``~/.config/.ewatercycle/ewatercycle.yaml``
```

ewatercycle/config/_validated_config.py
sverhoeven: Ok, thanks so paths are not checked for existence. Having typos in keys, missing keys all produce expected warnings and values.

sverhoeven (approved):
Thanks for greatly expanding the documentation. That will make it much easier to use.

Just a couple of corrections for docs and sdist packaging.

Peter9192:
> @Peter9192 Documentation is failing with `ModuleNotFoundError: No module named 'ruamel'`. I cannot reproduce this error locally, and I already tried a few things. Since you did some debugging with `pydoc_auto_imports`, do you have any idea where to look?

I compared with ESMValTool.
- They don't have ruamel in their mocked imports. Removing it, however, didn't have any effect.
- They have a separate requirements.txt in the docs folder. Adding it there resolves the problem. However, it does duplicate our dependencies again, which is exactly opposite of what we were trying to achieve.
- An alternative method would be to go back to manually specifying how RTD should build the package using a [`.readthedocs.yaml`](https://docs.readthedocs.io/en/stable/config-file/v2.html#packages) file. However, the 'install requires esmvaltool" in setup.py is making this option a PITA.

Feel free to keep or revert my last commit if you (dis)like it.

PR #49, opened 2021-04-13
Deduplicate dependencies in environment.yml

stefsmeets:
Hi all, this PR removes the duplicate dependencies from `environment.yml`. Dependencies that can be installed using pip should be added to `setup.py`. 

EDIT: I tried setting the python version in `environment.yml` to 3.9, because this greatly reduces the time spent resolving the environment (from at least the time it takes to push this branch and create this PR down to a few seconds). However, this causes all the tests to fail so it is better done in a separate PR.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,5 +5,5 @@ channels:
   - esmvalgroup
   - defaults
 dependencies:
-  - python>=3.7
+  - python>=3.9
   - esmvaltool-python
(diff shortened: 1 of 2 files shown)


stefsmeets:
Setting the python version to 3.9 is probably better done in a separate PR, because it requires messing with sonarcloud/github actions 😅

sverhoeven (approved):
Nice, I was able to run the unit-tests in a fresh environment. 

Good to see the dependencies sorted.

PR #54, opened 2021-04-15
Plot hydrograph

stefsmeets:
Hi everyone, this PR adds a function to plot a hydrograph.

Documentation: https://ewatercycle--54.org.readthedocs.build/en/54/apidocs/ewatercycle.analysis.html

## Example

```python
# Generate some data

import pandas as pd
import numpy as np

ntime = 300

dti = pd.date_range("2018-01-01", periods=ntime, freq="d")

discharge = {
    'discharge_a': pd.Series(np.linspace(0, 2, ntime), index=dti),
    'discharge_b': pd.Series(3*np.random.random(ntime)**2, index=dti),
    'discharge_c': pd.Series(2*np.random.random(ntime)**2, index=dti),
    'reference': pd.Series(np.random.random(ntime)**2, index=dti),
}

df = pd.DataFrame(discharge)

precipitation = {
    'precipitation_a': pd.Series(np.random.random(ntime)/20, index=dti),
    'precipitation_b': pd.Series(np.random.random(ntime)/30, index=dti),
}

df_pr = pd.DataFrame(precipitation)

# Plot hydrograph

from ewatercycle.analysis import hydrograph
hydrograph(df, reference='reference', precipitation=df_pr)
```

Output:

image


stefsmeets:
Hi @Peter9192 @SarahAlidoost , this PR is ready for review. Could one of you have a look, and let me know what you think?

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.dates import DateFormatter
+
+
+def hydrograph(
+    discharge: pd.DataFrame,
+    *,
+    reference: str,
+    precipitation: pd.DataFrame = None,
+    fname: Union[os.PathLike, str] = None,
```
Peter9192: I'm not sure if this is necessary. I'm fine with keeping it, but maybe as the last optional argument? Because I think the others are more important. Perhaps also use "save_as" instead of  "fname"? Or just "filename"

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    dpi: int = None,
```
Peter9192, line 14: Perhaps consider passsing **matplotlib_kwargs?

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    discharge_cols = discharge.columns.drop('reference')
+    y_obs = discharge[reference]
+    y_sim = discharge[discharge_cols]
+
+    fig, (ax, ax_tbl) = plt.subplots(
+        nrows=2,
+        ncols=1,
+        dpi=dpi,
+        figsize=(10, 10),
```
Peter9192: This is something I'd like to be able to tinker with as a user as well

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+        gridspec_kw={'height_ratios': [3, 1]},
```
Peter9192, line 71: Doesn't this depend on the number of datasets you need to list in the table? I guess it's fine for now, but just not sure if it makes sense to hardcode it.

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    handles, labels = ax.get_legend_handles_labels()
+
+    # Add precipitation as bar plot to the top if specified
+    if precipitation is not None:
+        ax_pr = ax.twinx()
+        ax_pr.invert_yaxis()
+        ax_pr.set_ylabel(f'Precipitation ({precipitation_units})')
+
+        for pr_label, pr_timeseries in precipitation.iteritems():
+            ax_pr.bar(
```
Peter9192, line 91: Ideally, if precip data is from the same source as discharge data, it would be nice if they had the same color. Otherwise, maybe explicitly set a completely different color cycle, so there is no suggestion of a relationship between similarly colored discharge/precip.

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+                pr_timeseries.index.values,
+                pr_timeseries.values,
+                alpha=0.4,
+                label=pr_label,
+            )
+
+        # tweak ylim to make space at bottom and top
+        ax_pr.set_ylim(ax_pr.get_ylim()[0] * (7 / 2), 0)
+        ax.set_ylim(0, ax.get_ylim()[1] * (7 / 5))
```
Peter9192, line 101: Isn't this very much dependent on the input data?

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # prepend handles/labels so they appear at the top
+        handles_pr, labels_pr = ax_pr.get_legend_handles_labels()
+        handles = handles_pr + handles
+        labels = labels_pr + labels
+
+    # Put the legend outside the plot, at the bottom
+    ax.legend(handles, labels, bbox_to_anchor=(1.10, 1), loc='upper left')
```
Peter9192: "bottom" and "upper left" are contradictory?

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+    ax_tbl.set_axis_off()
+
+    # give more vertical space in cells
+    table.scale(1, 1.5)
+
+    if fname is not None:
+        fig.savefig(fname, bbox_inches='tight', dpi=dpi)
+
+    return ax
```
Peter9192: Why return only the figure axes and not the table? Consider returning the figure instance as well?

tests/test_analysis.py
```diff
@@ -0,0 +1,36 @@
+from ewatercycle.analysis import hydrograph
+from matplotlib.testing.decorators import image_comparison
+import pandas as pd
+import numpy as np
+
+
+@image_comparison(
```
Peter9192, line 7: That's cool! I didn't know about this :-)

Peter9192 (changes requested):
Hey @stefsmeets , nicely done! In addition to my suggestions below, I think it would also be nice to give the user control over the colors used in the figure (e.g. to make them consistent throughout a paper).

ewatercycle/analysis/__init__.py
stefsmeets: From your comment I don't understand what is or is not necessary. Can you clarify? Or is this only about the name of filename argument? I like `save_as` tbh.

ewatercycle/analysis/__init__.py
stefsmeets: Yeah, I was considering that. There are 2 different plots in one, the line plot (discharge) and the bar plot (observations), so this may end up being somewhat/half-complete implementation.

Did you have any specific use cases in mind?

ewatercycle/analysis/__init__.py
stefsmeets: What do you want to tinker with? The number of rows/columns? `figsize`?

ewatercycle/analysis/__init__.py
stefsmeets: It's all green to me, I cannot make up from your comment what it is that you mean.

ewatercycle/analysis/__init__.py
stefsmeets: No, this solution makes it independent of the input data. This ensures that the discharges takes up the bottom 5/7th and the precipitation the top 2/7th, which is one of the solutions I took from the comparison notebook. It does sort of rely on the matplotlib ylim scaler to do its job.

An alternative solution would be to calculate the ylim using the min/max of the data.

ewatercycle/analysis/__init__.py
stefsmeets: My bad, I had it first at the bottom but forgot to update the comment!

ewatercycle/analysis/__init__.py
stefsmeets: Yeah, better to return just the `fig`, since it contains both axes 👍

ewatercycle/analysis/__init__.py
stefsmeets: The first is really difficult to do, unless we explicitly pass some sort of mapping between the discharge/precipitation data. So I will re-use the color cycle from the discharge data to ensure the colors are different.

stefsmeets: Alternatively, we could pass a dict which maps the datasets to user-defined colors.

ewatercycle/analysis/__init__.py
Peter9192: I'm not sure if we need to include a save option in the hydrograph function. My personal preference is to be able to do this myself, so I'd rather have the function return the figure instance. I'm fine with having a save option for convenience, but then I think it should be the last keyword argument.

ewatercycle/analysis/__init__.py
Peter9192: e.g. setting the colors, figsize, transparency

ewatercycle/analysis/__init__.py
Peter9192: yes, `figsize`

ewatercycle/analysis/__init__.py
Peter9192: My comments always refer to the last line. I guess gh decided to print the preceding lines to give you some context, but apparently this is confusing rather than helpful.

If you have many rows in the table, the aspect ratio may no longer be suitable. I'm just trying to imagine edge cases where these hard-coded layout settings may fail.

ewatercycle/analysis/__init__.py
Peter9192: Or, if the discharge and precip dataframes have the same column names, they could map to the same colors?

ewatercycle/analysis/__init__.py
Peter9192: Okay, now I see how it works. Cool, thanks!

ewatercycle/analysis/__init__.py
stefsmeets: I'm not sure if there is a simple way to do this in the current function. Any more tinkering and we may be better off rewriting the entire thing or implement it as a class. If you agree, I will make how to color the lines a new issue, so we can tackle it in a follow-up PR.

Peter9192: Reusing the color cycle is acceptable for now 👍

ewatercycle/analysis/__init__.py
stefsmeets: This implementation performs well with up to about 7. Any higher and you start to get some overlap with the x labels, and after 10 it overlaps with the data. We said it should work well with 3. Do you think we should support more than 7 in this PR?

ewatercycle/analysis/__init__.py
stefsmeets: Personal preference aside, having a some sort `filename` argument is what we agreed on when we defined what this function should look like 😅

ewatercycle/analysis/__init__.py
stefsmeets: This can now be solved using the `figsize` argument if the height needs to be increased to make room for the table.

ewatercycle/analysis/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+        With, height of the plot in inches.
+    **kwargs:
+        Options to pass to the matplotlib plotting function
+
+    Returns
+    -------
+    fig : `matplotlib.figure.Figure`
+    ax, ax_tbl : tuple of `matplotlib.axes.Axes`
+    """
+    discharge_cols = discharge.columns.drop('reference')
```
SarahAlidoost: There is a bug in this line: `reference` should not be string `'reference'` in `.drop(reference)`.

stefsmeets: Great catch, thanks @SarahAlidoost ! 🕵️‍♀️

ewatercycle/analysis/__init__.py
Peter9192: Okay, I missed that one, sorry. Anyway, can we still put it last?

SarahAlidoost (commented):
@stefsmeets thank you. I could run this branch on some data. The hydrograph looks good to me and the calculated metrics are correct. Could you please have a look at the github actions which are red?

stefsmeets:
> @stefsmeets thank you. I could run this branch on some data. The hydrograph looks good to me and the calculated metrics are correct. Could you please have a look at the github actions which are red?

Thanks @SarahAlidoost , I was just working on this. Should be fixed now 🤞

Peter9192 (approved):
Thanks for addressing my comments @stefsmeets ! Looks great!

PR #55, opened 2021-04-15
Add PCRGlobWB model class

Peter9192:
This PR adds PCRGlobWB to `ewatercycle.models`. It includes an example notebook. Usage is as follows:

```python
from ewatercycle.models import PCRGlobWB
model = PCRGlobWB()
cfg_file, cfg_dir = model.setup(
    input_dir='./pcrglobwb_example_case/', 
    cfg_file='./pcrglobwb_example_case/setup.ini',
    additional_input_dirs=[],  # for specifying additional data paths that the model needs access to
    **kwargs,  # for overriding settings in `cfg_file`, e.g. meteoOptions = {"temperatureNC": "era5_tas_1990_2000.nc"}
)
model.initialize(cfg_file) 
```

The `cfg_dir` returned by `model.setup(...)` is the newly created working directory for this specific experiment. It is independent of `input_dir`. 

Passing custom forcing data is not yet included.


ewatercycle/config/_validators.py
```diff
@@ -138,4 +138,6 @@ def validate_path(value, allow_none=False):
     # wflow specific
     'wflow.singularity_image': validate_string_or_none,
     'wflow.docker_image': validate_string_or_none,
+    'pcrglobwb.singularity_image': validate_string_or_none,
```
nielsdrost, line 141: the default ewatercycle.yaml file (also in this package) does not contain the new keys, by default I get a "key error".

ewatercycle/models/pcrglobwb.py
```diff
@@ -0,0 +1,152 @@
...
+class PCRGlobWB(AbstractModel):
+    """eWaterCycle implementation of PCRGlobWB hydrological model.
+
+    Attributes
+        bmi (Bmi): GRPC4BMI Basic Modeling Interface object
+    """
+    def setup(  # type: ignore
+            self,
+            input_dir: PathLike,
+            cfg_file: PathLike,
```
nielsdrost, line 27: should this be cfg_file_template?

nielsdrost (commented):
Awesome work @Peter9192 ! Some small comments, otherwise good to go I think!

ewatercycle/config/_validators.py
Peter9192: Good point! Done

ewatercycle/models/pcrglobwb.py
Peter9192: I think in most cases users will just pass their own cfg_file, which they don't consider as template, but rather as the *actual* cfg they want to use. Also this is more in line with PyMT which allows you to pass a cfg_file as well. Once we have a nice and complete template, we should set this as the default (so no need to pass a template via setup), and whatever the user passes as cfg_file just overrides the defaults.

So I think cfg_file is fine, but if you insist I will change it.

sverhoeven: Where did you see in pymt that cfg_file can be passed to setup? I only see path which is same as our work_dir arg.

```python
from pymt.models import Hydrotrend

??Hydrotrend.setup
Signature: Hydrotrend.setup(self, *args, **kwds)
Source:   
    def setup(self, *args, **kwds):
        """Set up a simulation.

        Parameters
        ----------
        path : str, optional
            Path to a folder to set up the simulation. If not given,
            use a temporary folder.

        Returns
        -------
        str
            Path to the folder that contains the set up simulation.
        """
        if len(args) == 0:
            dir_ = tempfile.mkdtemp()
        else:
            dir_ = args[0]

        self._parameters.update(kwds)

        FileSystemLoader(self.datadir).stage_all(dir_, **self._parameters)

        config = self._meta.run["config_file"]
        if config["contents"] and not config["path"]:
            config["path"] = dir_

        if config["path"]:
            with as_cwd(dir_):
                config_file = FileTemplate.write(
                    config["contents"], config["path"], **self._parameters
                )
        else:
            config_file = None

        return config_file, os.path.abspath(dir_)
File:      ~/miniconda38/envs/pymt/lib/python3.9/site-packages/pymt/framework/bmi_setup.py
Type:      function
```

Peter9192: @sverhoeven Check, you're right. It says "Here, we didn’t use the outputs from setup because the run directory has been specified, and the configuration file is created within it." https://pymt.readthedocs.io/en/latest/usage.html#model-setup

nielsdrost: Thanks @sverhoeven! In any case I propose we let this be for now. Let's plan a design session to come up with a good way to do this in general.

ewatercycle/config/_validators.py
nielsdrost: Thanks!

nielsdrost:
Thanks @Peter9192! Not sure what's up with the sonarcloud analysis. Lets look into this if the problem occurs more often.

PR #56, opened 2021-04-16
Add marrmot M01 model class

SarahAlidoost:


ewatercycle/config/ewatercycle.yaml
```diff
@@ -5,7 +5,9 @@ singularity_dir: null
 output_dir: null
 wflow.singularity_image: null
 wflow.docker_image: null
+marrmot.singularity_image: null
+marrmot.docker_image: null
 lisflood.singularity_image: null
 lisflood.docker_image: null
 pcrglobwb.singularity_image: null
-pcrglobwb.docker_image: null
+pcrglobwb.docker_image: null
```
nielsdrost:
```suggestion
pcrglobwb.docker_image: null

```

ewatercycle/models/marrmot.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def _generate_work_dir(work_dir: PathLike = None):
+    """
+    Args:
+        work_dir: If work dir is None then create sub-directory in CFG['output_dir']
+    """
+    if work_dir is None:
+        scratch_dir = CFG['output_dir']
+        # TODO this timestamp isnot safe for parallel processing
```
nielsdrost, line 37: This is going to be a problem quickly. Perhaps add milliseconds to kick the can down the road?

nielsdrost (approved):
Nice work @SarahAlidoost  and @sverhoeven !

The `get_value_as_xarray` function seems to only work on flux_out_Q. With both P and T it crashes. Is this "future work"? :-)

Other than that good to go, so feel free to merge (perhaps add an issue on the limited use of get_value_as_xarray)

nielsdrost:
Sonarcloud is complaining about duplicate lines in test code. I guess we should either fix or ignore test code in quality verification. @sverhoeven do you know what the approach usually is?

SarahAlidoost:
> Nice work @SarahAlidoost and @sverhoeven !
> 
> The `get_value_as_xarray` function seems to only work on flux_out_Q. With both P and T it crashes. Is this "future work"? :-)
> 
> Other than that good to go, so feel free to merge (perhaps add an issue on the limited use of get_value_as_xarray)

Thank you @nielsdrost. I made issue eWaterCycle/ewatercycle-marrmot#6 and added a TODO and `NotImplementedError` to the function.

ewatercycle/models/marrmot.py
SarahAlidoost: When I was working with marrmot calibration and parallel processing, I found that this solution does not work because the CPU time would be the same for all cores. We added a random filename generator there.

sverhoeven:
> Sonarcloud is complaining about duplicate lines in test code. I guess we should either fix or ignore test code in quality verification. @sverhoeven do you know what the approach usually is?

Sonarcloud does not report duplication in test code only in [ewatercycle/models/marrmot.py](https://sonarcloud.io/component_measures?id=eWaterCycle_ewatercycle&metric=new_duplicated_lines_density&pullRequest=56&view=list). I would not change the gate, but do some work after EGU to lower code duplication and increase coverage.

The get_value_as_xarray method should be moved to a super class, as all models we currently have can use same method. However later we will get some models that will not be able to return xarray due to their grid type. So we might need a sub class of AbstractModel to handle implement get_value_as_xarray and the models can be subclasses of that one.

ewatercycle/models/marrmot.py
sverhoeven: Created https://github.com/eWaterCycle/ewatercycle/issues/67

PR #65, opened 2021-04-20
Add an example preprocessing notebook

SarahAlidoost:
This notebooks generates forcing data for gridded models lisflood, wflow and pcrglob and plots the results. 
I want to use the output of this notebook for the presentation at EGU.


examples/preprocessing.ipynb
```diff
@@ -0,0 +1,721 @@
+{
```
nielsdrost, line 1: <p>Should we crop these logs from the notebook show it becomes a bit more readable?</p><p> </p><p>Something like:</p><p> </p><p>These tasks will be executed: diagnostic_daily/tasmin, diagnostic_daily/script, diagnostic_daily/rsds, diagnostic_daily/tasmax, diagnostic_daily/tdps, diagnostic_daily/tas, diagnostic_daily/vas, diagnostic_daily/pr, diagnostic_daily/uas </p><p>Running 9 tasks sequentially </p><p>Maximum memory used (estimate): 0.2 GB</p><p>Successfully completed tasks</p><p>Note: Log for forcing generation abbreviated for clarity</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/eWaterCycle/ewatercycle/pull/65/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

nielsdrost (approved):
Nice work @SarahAlidoost!

The only question I had was on the log output. Will you crop that here in this PR, or do you think it is better to do this locally when showing the notebook and/or creating a pdf?

examples/preprocessing.ipynb
SarahAlidoost: I made issue #68 for this.

SarahAlidoost:
> Nice work @SarahAlidoost!
> 
> The only question I had was on the log output. Will you crop that here in this PR, or do you think it is better to do this locally when showing the notebook and/or creating a pdf?

Thank you. I think it is better if I do it locally which is used only for creating slides/pdf.

PR #81, opened 2021-05-07
eWaterCycle Tutorial Notebook

nielsdrost:
This is a PR for a tutorial explaining how to use the eWaterCycle platform.

At the moment it does not work as the API is not fully implemented.


Peter9192:
Notebook was merged into docs in https://github.com/eWaterCycle/ewatercycle/pull/117. We'll continue to update it as we develop the API.

PR #82, opened 2021-05-10
Change how Wflow class is instantiated

### LINKED ISSUE #76: Bug: wflow model class: NoSectionError: No section: 'API'
If `config.ini` does not have section "API", the section should be added to the config, for example:

```python
cfg["API"] = { "RiverRunoff": "2, m^3/s"}
```

In wflow.py, the implementation at [this line](https://github.com/eWaterCycle/ewatercycle/blob/c5cb64dfe252c63fa2c8e3ad86cc963cfe4f4abe/ewatercycle/models/wflow.py#L65)  `cfg.set(section, option, value)` works only if the section already exists.

Peter9192:
- List versions: `ewatercycle.models.Wflow.available_versions`
- Add `WflowParameterSet`: 
- Forcing can be generated with ESMValTool, but also loaded from external source:
- Change the constructor: `ewatercycle.models.Wflow(version='2020.1.1', parameter_set=parameter_set, forcing=forcing)`
- Change the setup method
- Fewer configurable parameters

- Fixes #76


Peter9192:
Please wait with reviewing this, we're making some adaptations.

Peter9192:
Almost ready, just haven't finished testing the singularity option yet. I really like the `docker://....` syntax, but I'm not yet entirely sure how to assert there is a cached image available.

Peter9192:
Okay it seems to work now. Suggestions for better "pre-pull" options for singularity are welcome

ewatercycle/forcing/wflow.py
```diff
@@ -31,26 +31,25 @@ class WflowForcing(DefaultForcing):
     """Container for wflow forcing data."""
 
     # Model-specific attributes (ideally should have defaults):
-    netcdfinput: str = "inmaps.nc"
+    netcdfinput: Optional[str] = 'inmaps.nc'
```
bvreede:
```suggestion
    netcdfinput: Optional[str] = "inmaps.nc"
```

bvreede (approved):
I tested this PR with Docker together with @Peter9192. As setting up singularity on a mac was more than trivial, we postponed running the notebook with singularity; furthermore, there is an ongoing issue with (caching of) singularity containers, that we are working on in backlog item 1615 (azure, not github).

For now, this PR is OK with me!

Peter9192:
Thanks @bvreede

PR #83, opened 2021-05-10
Update marrmot01 class

SarahAlidoost:


ewatercycle/models/marrmot.py
```diff
@@ -42,42 +42,70 @@ def _generate_work_dir(work_dir: PathLike = None):
 
 
 class MarrmotM01(AbstractModel):
-    """eWaterCycle implementation of Marrmot Collie River 1 (traditional bucket) hydrological model.
+    """eWaterCycle implementation of Marrmot Collie River 1 (traditional bucket) hydrological model. It sets MarrmotM01 parameter with an initial value that is the mean value of the range specfied in `model parameter range file <https://github.com/wknoben/MARRMoT/blob/master/MARRMoT/Models/Parameter%20range%20files/m_01_collie1_1p_1s_parameter_ranges.m>`_.
```
Peter9192:
```suggestion
    """eWaterCycle implementation of Marrmot Collie River 1 (traditional bucket) hydrological model. 
    
    It sets MarrmotM01 parameter with an initial value that is the mean value of the range specfied in `model parameter range file <https://github.com/wknoben/MARRMoT/blob/master/MARRMoT/Models/Parameter%20range%20files/m_01_collie1_1p_1s_parameter_ranges.m>`_.
```

See https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

ewatercycle/models/marrmot.py
```diff
@@ -42,42 +42,70 @@ def _generate_work_dir(work_dir: PathLike = None):
...
         super().__init__()
+        self.version = version
+        if CFG['container_engine'].lower() == 'singularity':
+            self._set_singularity_image()
+        elif CFG['container_engine'].lower() == 'docker':
+            self._set_docker_image()
+        else:
+            raise ValueError(
+                f"Unknown container technology in CFG: {CFG['container_engine']}"
+            )
```
Peter9192: Why not set both? You now have repeated code (line 127 also has this same if-block). I know it's unlikely, but what if CFG is changed in between instantiation and starting the container?

ewatercycle/models/marrmot.py
```diff
@@ -173,7 +215,7 @@ def _create_marrmot_config(self) -> PathLike:
             store_ini=self.store_ini,
         )
 
-        config_file = self.work_dir / 'marrmot-m01_config.mat'
+        config_file = work_dir / Path('marrmot-m01_config.mat')
```
Peter9192, line 227:
```suggestion
        config_file = work_dir / 'marrmot-m01_config.mat'
```
if `work_dir` is a Path instance (as the type hints suggests), then this should just work I think.

ewatercycle/models/marrmot.py
```diff
@@ -42,42 +42,70 @@ def _generate_work_dir(work_dir: PathLike = None):
...
 
     # unable to subclass with more specialized arguments so ignore type
     def setup(self,  # type: ignore
-              forcing: Union[ForcingData, PathLike],
-              maximum_soil_moisture_storage: float = 1000.0,
-              initial_soil_moisture_storage: float = 900.0,
+              maximum_soil_moisture_storage: float = None,
+              initial_soil_moisture_storage: float = None,
               start_time: datetime = None,
               end_time: datetime = None,
```
Peter9192: Start time and end time are not listed under `model.parameters`. I would expect the output of that property to match the input arguments for setup.

Peter9192: It would also be helpful if they can be passed as strings instead of datetime objects. So if `model.parameters` provides  `(start_time, "20210101")`, then `model.setup(start_time="20210102")` might be more convenient than `start_time = datetime.Datetime(2020, 1, 1))`

tests/models/test_marrmotm01.py
```diff
@@ -47,12 +27,27 @@ def forcing_file(self):
...
+        if m.bmi:
+            # Clean up container
+            del m.bmi
+
+    @pytest.fixture
+    def model_with_setup(self, model: MarrmotM01):
+        cfg_file, cfg_dir = model.setup()
         return model, cfg_file, cfg_dir
 
+    def test_parameters(self, model, forcing_file):
```
Peter9192, line 42: Are start_time and end_time considered parameters as well? Since they're part of the `setup` function, I'm inclined to see them as such.

tests/models/test_marrmotm01.py
```diff
@@ -101,9 +94,76 @@ def test_get_value_as_xarray(self, model_with_setup):
...
+
+
+class TestWithDatesOutsideRangeSetupAndExampleData:
+    @pytest.fixture
+    def forcing_file(self):
+        # Downloaded from
+        # https://github.com/wknoben/MARRMoT/blob/master/BMI/Config/BMI_testcase_m01_BuffaloRiver_TN_USA.mat
+        return Path(__file__).parent / 'data' / 'BMI_testcase_m01_BuffaloRiver_TN_USA.mat'
+
+    @pytest.fixture
```
Peter9192, line 153: Do I understand it correctly that this fixture is specific to the class it belongs to?

tests/models/test_marrmotm01.py
```diff
@@ -101,9 +94,76 @@ def test_get_value_as_xarray(self, model_with_setup):
...
+    def model_with_setup(self, model: MarrmotM01):
+        cfg_file, cfg_dir = model.setup(
+            maximum_soil_moisture_storage=1234,
+            initial_soil_moisture_storage=4321,
+            start_time=datetime(1990, 1, 1, tzinfo=timezone.utc),
+            end_time=datetime(1991, 12, 31, tzinfo=timezone.utc),
+        )
+        return model, cfg_file, cfg_dir
+
+    def test_setup(self, model_with_setup):
```
Peter9192, line 133: So this `model_with_setup` refers to `self.model_with_setup`?

Peter9192 (changes requested):
Looks good, and seems to be working as well for what I've tried. I have a few small comments and questions.

Peter9192:
And on the notebook: maybe it would be nice to execute a few time steps to see how to actually run the model.

tests/models/test_marrmotm01.py
sverhoeven: correct, model_with_setup is a fixture scoped to the class it is in

tests/models/test_marrmotm01.py
sverhoeven: yep, I used classes to group tests. The fixtures those tests needed can also be scoped in the class

ewatercycle/models/marrmot.py
sverhoeven: We could use `self.forcing_start_time` for the default `start_time`. 

For datetime as string I would like to standardize on [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) format.
Also the forcing uses UTC while dates given setup could be timezone less, which should be handled elegantly.

ewatercycle/models/marrmot.py
SarahAlidoost: Right, I applied your suggestion. However, this means that the key 'singularity_dir' should exist in CFG. See the example notebook where ewatercycle.yaml is overwritten.

ewatercycle/models/marrmot.py
SarahAlidoost: First I applied your suggestion, but then I bring it back due to mypy error:  `Unsupported left operand type for / ("_PathLike[Any]")`

ewatercycle/models/marrmot.py
SarahAlidoost: now, the time of model in UTC and ISO format string can be given as 'YYYY-MM-DDTHH:MM:SS+00:00'. and start/end time are added to model.parameters.

tests/models/test_marrmotm01.py
SarahAlidoost: it is fixed.

SarahAlidoost:
> And on the notebook: maybe it would be nice to execute a few time steps to see how to actually run the model.

I added a while loop, please note that [case -1 notebook](https://github.com/eWaterCycle/technicalPaperExampleNotebooks/blob/main/Case-1_Marrmot_Merrimack_ESMValTool.ipynb) shows how to run the model, too.

ewatercycle/models/marrmot.py
sverhoeven: What happens when you do `model.setup(start_time="2021-01-15T00:00:00+05:00")` will mat file be set to 2021-01-15 or 2021-01-16? Can a test be added

SarahAlidoost:
> Looks good, and seems to be working as well for what I've tried. I have a few small comments and questions.

Thank you, I addressed your comments, please have another look.

ewatercycle/models/marrmot.py
Peter9192: That's strange. Can you try calling this (on the command line with you ewatercycle conda env activated)?

`mypy -c 'from pathlib import Path; Path(__file__).parent.parent / "foo" / "bar"'`

(from: https://gitter.im/python/typing?at=5d25171ff867ec1687ddc61a)

ewatercycle/models/marrmot.py
SarahAlidoost: The last part i.e. `+00:00` is in `HH:MM` and shows UTC offset. In your example, the offset is 5 hours and it returns `datetime.datetime(2021, 1, 15, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=18000)))`, and it is still 2021, 1, 15, 0, 0. 
I think we should consider the UTC time as a standard (like the time in forcing data) and add a function to check that `start_time` is either in this format 'YYYY-MM-DDTHH:MM:SS+00:00' or this 'YYYY-MM-DDTHH:MM:SS' , and also to inform the user that the given `start_time` should be a UTC time.

ewatercycle/models/marrmot.py
SarahAlidoost: Your command line does not return an error. It seems that there is a difference in using `from pathlib import Path` vs  `from os import PathLike`. You can try the lines below in a python script and run the mypy on that script. It reproduces the error that I mentioned. 

```python
from pathlib import Path
from os import PathLike

def pathlike_fun(work_dir: PathLike):
    config_file = work_dir / 'marrmot-m01_config.mat'
    return config_file

def path_fun(work_dir: Path):
    config_file = work_dir / 'marrmot-m01_config.mat'
    return config_file
```

Peter9192: Yep, then I also get the error. @sverhoeven what is the reason to use `PathLike` rather than `Path` or perhaps `PurePath`?

Peter9192 (approved):
Thanks @SarahAlidoost I'm happy with it as is, although the mypy error on PathLike things still annoys me.

ewatercycle/models/marrmot.py
sverhoeven: No particular reason, it was used in the rest of the code so I continued with it.

Peter9192: Haha, ok I did the same. Perhaps we can change it here @SarahAlidoost ?

ewatercycle/models/marrmot.py
sverhoeven: Agree, instead of looking at string if timezone == UTC we could use datetime.tzinfo

```
>>> from datetime import datetime, timezone, timedelta
>>> datetime(2021, 1, 15, 0, 0, tzinfo=timezone(timedelta(seconds=18000))).tzinfo == timezone.utc
False
>>> datetime.now(timezone.utc).tzinfo == timezone.utc
True
>>> datetime.now().tzinfo == timezone.utc
False
```

ewatercycle/models/marrmot.py
SarahAlidoost: I couldn't change it only in this part of the code, because mypy returned several `Incompatible types in assignment` errors in other parts. So, I skipped changing it for now.

ewatercycle/models/marrmot.py
SarahAlidoost: I changed the [datetime iso format] to a [datetime64 iso format] i.e. 'YYYY-MM-DDTHH:MM:SSZ' that is used by [numpy](https://numpy.org/doc/stable/reference/generated/numpy.datetime_as_string.html), and xarray. I also added a util function to parse and check if time is UTC. @sverhoeven please let me know if you are happy with the changes, then we merge the PR. Thanks.

sverhoeven (approved):
Looks good to me. Was able to run example notebook.

My editor raised some warnings which I fixed .
To test get_time I added some tests.

PR #85, opened 2021-05-17
Change API for PCRGlobWB model class

Peter9192:
- Added `PCRGlobWBParameterSet` and `PCRGlobWBForcing` classes
- Changed the constructor to accept `version`, `parameter_set` and `forcing` inputs
- Update the example/test notebook to reflect the updated API
- Reduce the number of configurable parameters and use an "ewatercycle" flavour instead of just exposing the model config.


ewatercycle/models/pcrglobwb.py
```diff
@@ -15,94 +17,175 @@
...
+
+
 class PCRGlobWB(AbstractModel):
     """eWaterCycle implementation of PCRGlobWB hydrological model.
 
-    Attributes
+    Args:
+
+        version: pick a version from :py:attr:`~available_versions`
+        parameter_set: instance of :py:class:`~PCRGlobWBParameterSet`.
```
sverhoeven, line 55: forcing argument is undocumented
```suggestion
        parameter_set: instance of :py:class:`~PCRGlobWBParameterSet`.
        forcing: a forcing data object. If not given uses forcing from parameter_set.
```

ewatercycle/models/pcrglobwb.py
```diff
@@ -15,94 +17,175 @@
...
         bmi (Bmi): GRPC4BMI Basic Modeling Interface object
     """
-    def setup(  # type: ignore
-            self,
-            input_dir: PathLike,
-            cfg_file: PathLike,
-            additional_input_dirs: Iterable[PathLike] = [],
-            **kwargs) -> Tuple[PathLike, PathLike]:
-        """Start model inside container and return config file and work dir.
+    available_versions = ('setters')
```
sverhoeven: Available version should be tuple of strings not a single string.
```suggestion
    available_versions = ('setters',)
```

ewatercycle/models/pcrglobwb.py
```diff
@@ -15,94 +17,175 @@
...
+                image=self.docker_image,
                 image_port=55555,
                 work_dir=str(self.work_dir),
                 input_dirs=[str(input_dir) for input_dir in input_dirs],
                 timeout=10,
             )
         elif CFG["container_engine"] == "singularity":
-            image = CFG["pcrglobwb.singularity_image"]
-
             message = f"No singularity image found at {image}"
```
sverhoeven: image var is not available 
```suggestion
            message = f"No singularity image found at {self.singularity_image}"
```

ewatercycle/models/pcrglobwb.py
```diff
@@ -15,94 +17,175 @@
...
-            assert Path(image).exists(), message
+            assert self.singularity_image.exists(), message
```
sverhoeven: This breaks when using `docker://<image name>`, can we remove this check?

ewatercycle/models/pcrglobwb.py
```diff
@@ -15,94 +17,175 @@
...
+        images = {
+            "setters": "ewatercycle/pcrg-grpc4bmi:setters",
+        }
+        self.docker_image = images[self.version]
 
-        return self.cfg_file, self.work_dir,
+    def _set_singularity_image(self):
+        # TODO auto detect sif file based on docker image and singularity dir.
+        images = {
+            "setters": "ewatercycle/pcrg-grpc4bmi:setters",
```
sverhoeven: This does not look like a correct image name for singularity. 
Can it be changed to a sif file or `docker://ewatercycle/pcrg-grpc4bmi:setters`?

sverhoeven (commented):
Started with reviewing.
I have some small inline suggestions. 
Still need to run notebook.

examples/pcrglobwb.ipynb
```diff
@@ -2,28 +2,28 @@
  "cells": [
```
sverhoeven, line 2: <p>I am getting</p><pre class="ql-syntax" spellcheck="false">/home/verhoes/git/eWaterCycle/ewatercycle/ewatercycle/models/pcrglobwb.py:48: UserWarning: No valid input path received for temperatureNC, defaulting to ''.
</pre><p> </p><p>I worked around it by doing </p><p> </p><pre class="ql-syntax" spellcheck="false">cp examples/pcrglobwb_example_case/forcing/temperature_2001to2010.nc&nbsp;&nbsp;examples/
</pre><p> </p><p>Can this be added as shell command or as instructions?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/eWaterCycle/ewatercycle/pull/85/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

sverhoeven, line 2: <p>Can we show the use of setup, by giving some arguments like</p><p> </p><pre class="ql-syntax" spellcheck="false">cfg_file, cfg_dir = pcrglob.setup(landSurfaceOptions={'debugWaterBalance': 'False'})
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/eWaterCycle/ewatercycle/pull/85/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='9'/>

ewatercycle/models/pcrglobwb.py
```diff
@@ -15,94 +17,175 @@
...
+        default_input_dir = self.parameter_set.input_dir
+
         for section, options in kwargs.items():
+            if not cfg.has_section(section):
+                cfg.add_section(section)
+
             for option, value in options.items():
-                cfg.set(section, option, value)
 
+                if Path(value).exists():
```
sverhoeven: On line 50 a value can be set to an empty string. Doing

```
In [1]: from pathlib import Path

In [4]: p = Path('')

In [5]: p.exists()
Out[5]: True

```

So when I run the notebook I get a pcrglobwb_ewatercycle.ini with
```
refETPotFileNC = /home/verhoes/git/eWaterCycle/ewatercycle/examples
```

I expected the refETPotFileNC to have kept the same value as the default config 
```
refETPotFileNC = forcing/CRU-TS3.21/merged_1958_to_2010/CRU-TS3.21_ERA-40_ERA-Interim_daily_referencePotET_1958_to_2010.nc

```

sverhoeven (changes requested):
Was able to run notebook with a extra manual command. Can command be added to notebook?

ewatercycle/models/pcrglobwb.py
Peter9192: Not sure how you triggered this, but I think it is solved now.

sverhoeven: By running examples/pcrglobwb.ipynb notebook until the `setup()` method. I then looked at output of `pcrglob.parameters`:

```
('meteoOptions',
  {'precipitationNC': '/home/verhoes/git/eWaterCycle/ewatercycle/examples/pcrglobwb_example_case/forcing/precipitation_2001to2010.nc',
   'temperatureNC': '/home/verhoes/git/eWaterCycle/ewatercycle/examples/temperature_2001to2010.nc',
   'referenceETPotMethod': 'Hamon',
   'refETPotFileNC': 'forcing/CRU-TS3.21/merged_1958_to_2010/CRU-TS3.21_ERA-40_ERA-Interim_daily_referencePotET_1958_to_2010.nc'}),
```

If I run the notebook again the `refETPotFileNC` value now is correct as it is unchanged from the default_config.

However the temperatureNC value is weird
Because you used 
```
forcing = ewatercycle.forcing.load_foreign(
    target_model = "pcrglobwb",
    start_time = "2001-01-01T00:00:00Z",
    end_time = "2010-12-31T00:00:00Z",
    directory = "./pcrglobwb_example_case/forcing",
    shape = None,
    forcing_info = dict(
        precipitationNC = "pcrglobwb_example_case/forcing/precipitation_2001to2010.nc",
        temperatureNC = "temperature_2001to2010.nc"
))
```

I expected value of temperatureNC to be `forcing.directory / "temperature_2001to2010.nc"` so after relativizing against `inputDir` I expected `forcing/temperature_2001to2010.nc`. However now it is set to  '/home/verhoes/git/eWaterCycle/ewatercycle/examples/temperature_2001to2010.nc' which is `./temperature_2001to2010.nc` relative to notebook and that file does not exist

ewatercycle/models/pcrglobwb.py
```diff
@@ -11,98 +11,154 @@
...
-        """
         self._setup_work_dir()
-        self._setup_config(cfg_file, input_dir, **kwargs)
-        self._start_container(input_dir, additional_input_dirs)
+        self._setup_default_config()
+
+        self._update_config(
+            meteoOptions={
+                "temperatureNC": forcing.temperatureNC,
+                "precipitationNC": forcing.precipitationNC,
```
sverhoeven: My other inline suggestion is probably caused here. We should append `forcing.directory`.

Something like
```suggestion
                "temperatureNC": str(Path(forcing.directory) / forcing.temperatureNC),
                "precipitationNC": str(Path(forcing.directory) / forcing.precipitationNC),
```

sverhoeven (changes requested):
The forcing paths are not correctly resolved into the config file. Could you have a look at that?

ewatercycle/models/pcrglobwb.py
Peter9192: Right! Thanks for the elaborate feedback :-)

examples/pcrglobwb.ipynb
Peter9192: This is an advanced use cases that I decided to remove from the notebook.

Peter9192: Done, but I drastically reduced the configurable parameters (had discussed this with Niels a while ago), and made them use the 'ewatercycle' format instead of just exposing the full model config file.

ewatercycle/models/pcrglobwb.py
Peter9192: Solved now. The current scheme:
- User supplies `directory` and `temperatureNC` (relative to `directory`; clearly documented) in constructing the forcing object
- The model class will combine them and then expand into an absolute path
- That path is mounted on the container and substituted in the config file. 

There is still a way to configure non-default paths. `setup` now supports an "additional_dirs" argument, which allows the user to add any deviating paths in the config file, and to make sure the container is still able to find them, but that's an advanced use case that we should probably not advertise.

Peter9192:
@sverhoeven made some more changes, a.o. I reduced the number of configurable parameters (discussed with Niels a while ago). Could I bother you with another round?

ewatercycle/models/pcrglobwb.py
```diff
@@ -11,112 +11,195 @@
...
             self,
-            input_dir: PathLike,
-            cfg_file: PathLike,
-            additional_input_dirs: Iterable[PathLike] = [],
+            additional_input_dirs: Optional[Iterable[str]] = None,
             **kwargs) -> Tuple[PathLike, PathLike]:
         """Start model inside container and return config file and work dir.
 
         Args:
+            additional_input_dirs: By default, the model only has access to
```
sverhoeven: Do we want to expose this argument to the user now?  For me it is not needed.
Any extra file could be added to work_dir and pointed to in cfg_file. 
User will need to do a copy, but I think that is a big enough escape hatch.

ewatercycle/models/pcrglobwb.py
```diff
@@ -11,112 +11,195 @@
...
+                its working directory, the parameter sets, and the forcing
+                directory. This option makes it possible to add other external
+                inputs, that can then be configured in the config file.
+            **kwargs: Use :py:meth:parameters to see all configurable options for this model,
```
sverhoeven: Instead of pointing to parameters it would be nice to spell out the available arguments.
So have `start_time, end_time, routing_method, dynamic_flood_plain, max_spinups_in_years` as setup arguments with docstrings. Then it will be possible for example to document which routing methods are possible (accuTravelTime|kinematicWave) .
We can also do this later.

Fix link in Sphinx
```suggestion
            **kwargs: Use :py:meth:`parameters` to see all configurable options for this model,
```

ewatercycle/models/pcrglobwb.py
```diff
@@ -11,112 +11,195 @@
...
 
         if CFG["container_engine"] == "docker":
             self.bmi = BmiClientDocker(
-                image=CFG["pcrglobwb.docker_image"],
+                image=self.docker_image,
                 image_port=55555,
                 work_dir=str(self.work_dir),
-                input_dirs=[str(input_dir) for input_dir in input_dirs],
+                input_dirs=additional_input_dirs,
                 timeout=10,
```
sverhoeven: Similar to singularity if the Docker image has not been downloaded yet. It could take more than 10 seconds to download. 

We should have a similar try/catch as on line 192 or remove the timeout argument or fix https://github.com/eWaterCycle/grpc4bmi/issues/100 and https://github.com/eWaterCycle/grpc4bmi/issues/95

ewatercycle/models/pcrglobwb.py
```diff
@@ -11,112 +11,195 @@
...
+                    image=f"docker://{self.docker_image}",
+                    work_dir=str(self.work_dir),
+                    input_dirs=additional_input_dirs,
+                    timeout=15,
+                )
+            except FutureTimeoutError:
+                raise ValueError(
+                    "Couldn't spawn the singularity container within allocated"
+                    " time limit (15 seconds). You may try building it with "
+                    f"`!singularity run docker://{self.docker_image}` and try "
```
sverhoeven: When I got this exception in the example notebook, the building of the Singularity image kept going (as evident from the Jupyter log). So I let that complete and then ran setup again.

The `singularity run docker://ewatercycle/pcrg-grpc4bmi:setters` will run forever as it starts a BMI server. My suggestion would be to run `singularity exec docker://ewatercycle/pcrg-grpc4bmi:setters run-bmi-server -h` as it will show some help and then immediately exit.
Also having a `!` assumes that the error was thrown in a notebook. I don't think its wise to presume that and would remove the `!`.

The last sentence assumes the system admin has diverted the SINGULARITY_CACHEDIR to a shared location instead of the default `~/.singularity`. This is not possible according to https://singularity.hpcng.org/user-docs/3.7/build_env.html#sec-cache

sverhoeven (changes requested):
Notebook works.

If have some inline questions/suggestions

ewatercycle/models/pcrglobwb.py
Peter9192: Removed it 👍

ewatercycle/models/pcrglobwb.py
Peter9192: Let's do that later, since I'm not sure what to document for these options, and this is probably not the definitive set anyway.

ewatercycle/models/pcrglobwb.py
Peter9192: Aii, that's a good point. Maybe we cannot use the `docker://` syntax after all then, if we want the admin to manage the containers on the system. 

For now I've wrapped the whole container startup in a try/except block and adapted the message.

ewatercycle/models/pcrglobwb.py
sverhoeven: Later at https://dev.azure.com/nlesc-team-beta/Team%20Beta/_boards/board/t/Team%20Beta/Backlog%20items/?workitem=1625

ewatercycle/models/pcrglobwb.py
sverhoeven: Nice

ewatercycle/models/pcrglobwb.py
Peter9192: https://github.com/eWaterCycle/ewatercycle/issues/119

sverhoeven (approved):
LGTM

PR #87, opened 2021-05-26
Refactor forcing module

Peter9192:
Within the forcing module, there is now one file per model, with one class for each model (and a default): `DefaultForcing`, `WflowForcing`, `MarrmotForcing`, ... This makes it easy to add forcing-related code for new models. 

Each model implements a classmethod called `generate()` which serves as an secondary constructor. This makes it easy to either load existing data, or generate it on the fly:

```python
ewatercycle.forcing.generate(target_model=..., **kwargs)
ewatercycle.forcing.load(target_model=..., directory=..., **kwargs)
ewatercycle.forcing.reload(directory=...)  # will look for `ewatercycle_forcing.yaml` and use that
```

Generic options  for load/generate are defined in the `DefaultForcing`. Model-specific options are documented through the docstrings of the individual model classes. 

Since these objects are mostly containers of information, I used `dataclasses` to reduce boilerplate.

TODO 
- [ ] refactor tests
- [ ] update model classes that already use forcing
- [ ] standardize default options, time conventions etc.
- [ ] figure out how to disable [duplicated docstrings](https://stackoverflow.com/questions/51125415/how-do-i-document-a-constructor-for-a-class-using-python-dataclasses)


sverhoeven (approved):
Let merge this so we can work in parallel again

PR #84, opened 2021-05-17
Update lisflood class

SarahAlidoost:
Related issue #92

PR #89, opened 2021-05-31
Added tests for forcing.marrmot

sverhoeven:


tests/forcing/test_marrmot.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+    def mock_run(self):
+        """Store recipe for inspection and return dummy output."""
+        nonlocal data
+        data['data_during_run'] = self.data
+        return {'diagnostic_daily/script': MockTaskOutput()}
+
+    monkeypatch.setattr(Recipe, "run", mock_run)
+    return data
```
Peter9192: Cool, I really like the way this stores the data inside the fixture. I'm just not sure I understand where it is that this `data` is returned to?

tests/forcing/test_marrmot.py
```diff
@@ -0,0 +1,123 @@
...
+        del actual['preprocessors']['daily']['extract_shape']['shapefile']
+
+        assert actual == reference_recipe
+        assert actual_shapefile == sample_shape
+        assert 'MARRMoT' in actual_desc
+
+    def test_saved_yaml(self, forcing, tmp_path):
+        saved_forcing = load(tmp_path)
+
+        assert forcing == saved_forcing
```
Peter9192, line 125: Should we also add a test for the load_external function?

ewatercycle/forcing/__init__.py
```diff
@@ -39,8 +39,9 @@ def generate(target_model: str,
     if constructor is None:
         raise NotImplementedError(f'Target model `{target_model}` is not supported by the eWatercycle forcing generator')
     if model_specific_options is None:
-        return constructor.generate(dataset, start_time, end_time, shape)
-    forcing_info = constructor.generate(dataset, start_time, end_time, shape, **model_specific_options)
+        forcing_info = constructor.generate(dataset, start_time, end_time, shape)
+    else:
+        forcing_info = constructor.generate(dataset, start_time, end_time, shape, **model_specific_options)
```
Peter9192: Can we get rid of this conditional block if we make `model_specific_options` default to an empty dict instead of `None`?

Peter9192 (changes requested):

Peter9192:
Nice @sverhoeven, works fine, just a few comments/questions

ewatercycle/forcing/__init__.py
sverhoeven: Empty dict is mutable, which is not allowed as a default value for a argument. As it makes the value tied to the class instead of the object. Can make it simpler

tests/forcing/test_marrmot.py
sverhoeven: On line 36 I return the data dict. Because `mock_recipe_run` is a fixture. When used in test as an argument the value will be equal to data. 

`recipe.run()` is called in the test so I had to scope data to nonlocal so I could store the arguments of the call to `run()`.
I could not do `data = self.data` because then in the test, the value of the fixture would still point to the line 21

tests/forcing/test_marrmot.py
sverhoeven: test added

tests/forcing/test_marrmot.py
sverhoeven: I renamed the outer `data` var to `recorder` to make it more clear what it does

Peter9192:
CI is not happy because the following files still import old forcing_data.
ERROR tests/models/test_abstract.py
ERROR tests/models/test_lisflood.py
ERROR tests/models/test_marrmotm01.py

See #92. #84 and #95 should fix those. 

I'm okay with merging this already.

Peter9192 (approved):

sverhoeven:
Thanks for reviewing.
The CI will be fixed in of the other PRs

Issue #48, opened 2021-04-13
Document how to add a new model


sverhoeven:
To add a new model we would need to 
* Create model as subclass of AbstractModel (ewatercycle/models/abstract.py)
* Import model in `ewatercycle/models/__init__.py`
* Add  `ewatercycle/forcing/<model>.py`
* Register model in `ewatercycle/forcing/__init__.py:FORCING_CLASSES`
* Add model to docs/conf.py
* Write example notebook
* Write tests?
* If model needs custom parameter set class add it in `ewatercycle/parameter_sets/_<model name>.py`
* Add example parameter set in `ewatercycle/parameter_sets/__init__.py`
* Add container image to setup guide

Can the steps be documented?

sverhoeven:
List is now part of https://ewatercycle.readthedocs.io/en/latest/adding_models.html

PR #98, opened 2021-06-03
Bundle model-specific docstrings and hide their classes from public api

### LINKED ISSUE #97: Combining model-specific docstrings
We want to use a single entry point (i.e. `ewatercycle.forcing.load`) for instantiating different Forcing objects (i.e. they have different model-specific arguments). So

```python
ewatercycle.forcing.load(target_model='marrmot', start_time=..., ..., model_specific_options=None) 
>> MarrmotForcing(start_time=..., ...)

ewatercycle.forcing.load(target_model='wflow', start_time=..., ..., model_specific_options={Temperature: '/TEMP'} 
>> WflowForcing(start_time=..., ..., Temperature='/TEMP')
```

Since we want to push towards a uniform interface, we want to hide the model-specific classes from the documentation to discourage using them directly. The question is: how do we still provide good documentation of all model-specific options for the `load` function?

Peter9192:
closes #97

Look like this (and similar for `load_foreign`):

image

Docstrings are added to `forcing._model.ModelForcing.__init__` and `forcing._model.ModelForcing.generate`, and should adhere to the following format:

```python
def __init__(self, ...):
    """
        model_specific_option1: documentation
        model_specific_option2: documentation
    """

@classmethod
def generate(cls, ...):
    """
        None: no model-specific options needed for this model
    """
```

Those docstrings will be collected from all models, combined into one listing, and added to the docstrings of the `forcing.generate` and `forcing.load_foreign` methods. To render nicely, one also needs to add the model to `docs/conf.py`:
```python
# Nice formatting of model-specific input parameters
napoleon_custom_sections = [
    ('hype', 'params_style'),
    ('lisflood', 'params_style'),
    ('marrmot', 'params_style'),
    ('pcrglobwb', 'params_style'),
    ('wflow', 'params_style'),
]
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -207,6 +207,7 @@ autodoc_mock_imports = [
     'dask',
     'esmvalcore',
     'fiona',
+    'dateutil',
     'hydrostats',
     'matplotlib',
     'numpy',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -219,3 +219,12 @@ autodoc_mock_imports = [
 
 # Prevent alphabetic sorting of (@data)class attributes/methods
 autodoc_member_order = 'bysource'
+
+# Nice formatting of model-specific input parameters
+napoleon_custom_sections = [
+    ('hype', 'params_style'),
+    ('lisflood', 'params_style'),
+    ('marrmot', 'params_style'),
+    ('pcrglobwb', 'params_style'),
+    ('wflow', 'params_style'),
+]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -206,7 +206,7 @@ autodoc_mock_imports = [
     'cftime',
     'dask',
     'esmvalcore',
-    'fiona'
+    'fiona',
     'hydrostats',
     'matplotlib',
     'numpy',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -206,6 +206,7 @@ autodoc_mock_imports = [
     'cftime',
     'dask',
     'esmvalcore',
+    'fiona'
     'hydrostats',
     'matplotlib',
     'numpy',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -208,6 +208,7 @@ autodoc_mock_imports = [
     'esmvalcore',
     'fiona',
     'dateutil',
+    'shapely',
     'hydrostats',
     'matplotlib',
     'numpy',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -191,25 +191,6 @@ texinfo_documents = [
      'Miscellaneous'),
 ]
 
-autodoc_mock_imports = [
-    'basic_modeling_interface',
-    'cftime',
-    'dask',
-    'esmvalcore',
-    'fiona',
-    'dateutil',
-    'shapely',
-    'hydrostats',
-    'matplotlib',
-    'numpy',
-    'pandas',
-    'pyoos',
-    'grpc4bmi',
-    'ruamel.yaml',
-    'scipy',
-    'xarray',
-]
-
 # Prevent alphabetic sorting of (@data)class attributes/methods
 autodoc_member_order = 'bysource'
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -191,6 +191,25 @@ texinfo_documents = [
      'Miscellaneous'),
 ]
 
+autodoc_mock_imports = [
+    'basic_modeling_interface',
+    'cftime',
+    'dask',
+    'esmvalcore',
+    'fiona',
+    'dateutil',
+    'shapely',
+    'hydrostats',
+    'matplotlib',
+    'numpy',
+    'pandas',
+    'pyoos',
+    'grpc4bmi',
+    'ruamel.yaml',
+    'scipy',
+    'xarray',
+]
+
 # Prevent alphabetic sorting of (@data)class attributes/methods
 autodoc_member_order = 'bysource'
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,9 +19,16 @@
 import os
 import sys
 
+from ewatercycle.forcing import FORCING_CLASSES
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+from ewatercycle.version import __version__ as version
+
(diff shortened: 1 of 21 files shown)


ewatercycle/forcing/_lisflood.py
```diff
@@ -0,0 +1,110 @@
...
+    @classmethod
+    def generate(  # type: ignore
+        cls,
+        dataset: str,
+        start_time: str,
+        end_time: str,
+        shape: str,
+        extract_region: dict = None,
+    ) -> 'LisfloodForcing':
+        """
```
bouweandela, line 57: Maybe you could do something like:
```suggestion
        super().generate.__doc__ + """
```
to generate the complete documentation here?

ewatercycle/forcing/_lisflood.py
```diff
@@ -0,0 +1,110 @@
...
+    """Container for lisflood forcing data."""
+
+    # TODO check whether start/end time are same as in the files
+    def __init__(
+        self,
+        start_time: str,
+        end_time: str,
+        directory: str,
+        shape: str,
+        PrefixPrecipitation: Optional[str] = 'pr.nc',
```
bouweandela, line 23: You have probably discussed this at length already, but why are you using the weird capitalization? It would make a more uniform impression to stick with the PEP8 naming convention.

bouweandela:
I posted the two ideas I had on this topic, but please feel free to ignore them if it doesn't work for you.

ewatercycle/forcing/_lisflood.py
Peter9192: We're just exposing the model configuration settings. This is passed on directly to the lisflood configuration file that's automatically created. If we change it here, then there will be a mapping from 'ewatercycle model settings' to 'native hydrological model settings'. I think it's better to stick with the native model options for now.

docs/conf.py
```diff
@@ -219,3 +222,12 @@ def setup(app):
...
 autodoc_member_order = 'bysource'
+
+# Nice formatting of model-specific input parameters
+napoleon_custom_sections = [
+    ('hype', 'params_style'),
+    ('lisflood', 'params_style'),
+    ('marrmot', 'params_style'),
+    ('pcrglobwb', 'params_style'),
+    ('wflow', 'params_style'),
+]
```
sverhoeven, line 233: I like how it looks at https://ewatercycle--98.org.readthedocs.build/en/98/apidocs/ewatercycle.forcing.html#ewatercycle.forcing.generate
, but having an additional place to edit when adding a model is something I want to avoid.  Can we make it dynamic?
```suggestion
from ewatercycle.forcing import FORCING_CLASSES
napoleon_custom_sections = [(m, 'params_style') for m in FORCING_CLASSES.keys()]
```

ewatercycle/forcing/_hype.py
```diff
@@ -1,41 +1,35 @@
...
+        self,
+        start_time: str,
+        end_time: str,
+        directory: str,
+        shape: str,
+    ):
+        """
+            None: Hype does not have model-specific load options.
+        """
+        super().__init(start_time, end_time, directory, shape)
```
sverhoeven:
```suggestion
        super().__init__(start_time, end_time, directory, shape)
```
Same for other forcing classes

ewatercycle/forcing/_marrmot.py
```diff
@@ -1,27 +1,30 @@
...
-
-    # Model-specific attributes (preferably with default values):
-    forcing_file: str = 'marrmot.mat'
+    def __init__(
+        self,
+        start_time: str,
+        end_time: str,
+        directory: str,
+        shape: str,
+        forcing_file: Optional[str] = 'marrmot.mat',
```
sverhoeven, line 21: By not using dataclass we can make model specific options required.
```suggestion
        forcing_file: str,
    ):
```

Having defaults for load_foreign model_info felt to much as a surprise as the whole idea of load_foreign is to be specific.

sverhoeven (commented):
The rendered docs at https://ewatercycle--98.org.readthedocs.build/en/98/apidocs/ewatercycle.forcing.html# look nice. 

I however was thinking making it part of the Sphinx user docs not the generated API docs.
As I think hydrologists would rather read a tutorial or user manual then API docs.

Still lots of things to do, but the the current implementation is workable for me.

Peter9192:
> I however was thinking making it part of the Sphinx user docs not the generated API docs.
> As I think hydrologists would rather read a tutorial or user manual then API docs.

I agree with that, but I think the user docs should not replace the API docs, but rather complement them. I see that as one of the next steps.

ewatercycle/forcing/_marrmot.py
Peter9192: I removed it for marrmot. Do you also want to remove them for other models? Sometimes I think it does actually make sense to have defaults, and it saves typing. E.g. for wflow I like that the default variable names are the CMOR names.

docs/conf.py
Peter9192: It appears spinx doesn't like this import

sverhoeven: Hmmpf, tried to use .readthedocs.yml with conda install (17addd6e1fe7e219ae0338cefcdae76fc0105da6), but [build failed with `Command killed due to excessive memory consumption`](https://readthedocs.org/projects/ewatercycle/builds/13967669/)

I reverted it back to having a hardcoded list and updated #48

ewatercycle/forcing/_marrmot.py
sverhoeven: When the model has a single standard value then we can make that the default. Like for wflow var names.
Most filenames are not standardized as they include time ranges or can be in different formats.
So I think the default for filenames should be removed.

What do you think? Only keep non-file defaults

Peter9192: I'm fine with removing all defaults for filenames, but maybe we should do it in a separate PR. We still get issues in the tests from making `forcing_info` required while `shape` is not required in `__init__` (this allows for loading external forcing for which no shapefile is available). I'll make a new issue.

Peter9192: https://github.com/eWaterCycle/ewatercycle/issues/110

Peter9192:
All checks green. I opened #110 to continue the discussion about removing defaults. This PR is already big enough as it is...

PR #101, opened 2021-06-07
Lisflood forcing tests

sverhoeven:


tests/forcing/test_lisflood.py
```diff
@@ -0,0 +1,219 @@
...
+            'preprocessors': {'daily_radiation': {'convert_units': {'units': 'J m-2 '
+                                                                             'day-1'},
+                                                  'custom_order': True,
+                                                  'extract_region': {'end_latitude': 52.2,
+                                                                     'end_longitude': 11.9,
+                                                                     'start_latitude': 46.3,
+                                                                     'start_longitude': 4.1},
+                                                  'extract_shape': {'crop': True,
+                                                                    'method': 'contains',
+                                                                    'shapefile': '/home/verhoes/git/eWaterCycle/ewatercycle/tests/models/data/Rhine/Rhine.shp'},
```
SarahAlidoost: The test is failing because of this path. I suggest refactoring as:
```python 
f'{Path(__file__).parents[1]}/models/data/Rhine/Rhine.shp'
```
please implement this for other cases below and add `from pathlib import Path` too.

tests/forcing/test_lisflood.py
```diff
@@ -0,0 +1,219 @@
...
+            PrefixE0='lisflood_e0.nc',
+            PrefixES0='lisflood_es0.nc',
+            PrefixET0='lisflood_et0.nc'
+        )
+        assert forcing == expected
+
+    def test_recipe_configured(self, forcing, mock_recipe_run, reference_recipe, sample_shape):
+        actual = mock_recipe_run['data_during_run']
+        # Remove long description and absolute path so assert is easier
+        actual_desc = actual['documentation']['description']
```
SarahAlidoost, line 204: `actual_desc` is unused.

SarahAlidoost (changes requested):
@sverhoeven many thanks. I left a comment about refactoring the path of the shapefile. With the refactoring, the test passed on my local machine.

sverhoeven:
Having paths that only work on my computer is no good. Seems this branch needed a bit more work to be ready. Completed test so it also works on other computers.

Due to old forcing data CI still gives red. Could you try running test file locally with

```
pytest tests/forcing/test_lisflood.py
```

SarahAlidoost:
> Having paths that only work on my computer is no good. Seems this branch needed a bit more work to be ready. Completed test so it also works on other computers.
> 
> Due to old forcing data CI still gives red. Could you try running test file locally with
> 
> ```
> pytest tests/forcing/test_lisflood.py
> ```

Thank you for fixing the test. It is passed. I approved the PR, let's merge it.

SarahAlidoost (approved):

sverhoeven:
Thanks for reviewing

PR #102, opened 2021-06-09
Fixes mypy errors

sverhoeven:
Except for ewatercycle/forcing/wflow.py, those will be fixed in #82

### DIFF
--- a/ewatercycle/util.py
+++ b/ewatercycle/util.py
@@ -1,4 +1,3 @@
-from pathlib import Path
 from typing import Any, Tuple, Dict
 
 import fiona
--- a/ewatercycle/util.py
+++ b/ewatercycle/util.py
@@ -30,10 +30,10 @@ def var_to_xarray(model, variable):
 
     # Create xarray object
     da = xr.DataArray(data,
-                      coords = {'longitude': lon, 'latitude': lat, 'time': time},
-                      dims = ['latitude', 'longitude'],
-                      name = variable,
-                      attrs = {'units': model.get_var_units(variable)}
+                      coords={'longitude': lon, 'latitude': lat, 'time': time},
+                      dims=['latitude', 'longitude'],
+                      name=variable,
+                      attrs={'units': model.get_var_units(variable)}
                       )
 
     # Masked invalid values on return array:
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -58,6 +58,7 @@
             'pytest-cov',
             'pytest-mypy',
             'pytest-runner',
+            'types-python-dateutil',
```
Peter9192, line 61: Should we also mock this in docs/conf.py?

ewatercycle/util.py
```diff
@@ -71,12 +72,12 @@ def lat_lon_boundingbox_to_variable_indices(model, variable, latMin, latMax, lon
...
-    latModel = model.get_grid_x(model.get_var_grid(variable))
-    lonModel = model.get_grid_y(model.get_var_grid(variable))
-    nx = len(latModel)
+    lat_model = model.get_grid_x(model.get_var_grid(variable))
+    lon_model = model.get_grid_y(model.get_var_grid(variable))
+    nx = len(lat_model)
 
-    idx = [i for i,v in enumerate(latModel) if ((v > latMin) and (v < latMax))]
-    idy = [i for i,v in enumerate(lonModel) if ((v > lonMin) and (v < lonMax))]
+    idx = [i for i,v in enumerate(lat_model) if ((v > latMin) and (v < latMax))]
```
bvreede:
```suggestion
    idx = [i for i,v in enumerate(lat_model) if ((v > lat_min) and (v < lat_max))]
```

bvreede: should these variable names not also be adjusted?

ewatercycle/util.py
```diff
@@ -71,12 +72,12 @@ def lat_lon_boundingbox_to_variable_indices(model, variable, latMin, latMax, lon
...
+    idy = [i for i,v in enumerate(lon_model) if ((v > lonMin) and (v < lonMax))]
```
bvreede:
```suggestion
    idy = [i for i,v in enumerate(lon_model) if ((v > lon_min) and (v < lon_max))]
```

bvreede (commented):
FWIW: I checked this branch locally and did not find further mypy errors (except in `wflow.py`).

setup.py
sverhoeven: Not needed, build https://ewatercycle--102.org.readthedocs.build/en/102/ succeeded. Also we dont import that package, its something mypy needed.

ewatercycle/util.py
sverhoeven: Hmm, pycharm did not complain about those, but yes let's also adjust those vars.

PR #103, opened 2021-06-09
Dont store dir in forcing yaml

sverhoeven:


ewatercycle/forcing/default.py
```diff
@@ -37,10 +39,15 @@ def save(self):
...
+        target = Path(self.directory) / FORCING_YAML
+        # We want to make the yaml and its parent moveable,
+        # so the directory should not be included in the yaml file
+        inner_file = StringIO()
+        yaml.dump(self, inner_file)
+        inner = inner_file.getvalue()
+        inner = inner.replace(f'directory: {self.directory}\n', '')
         with open(target, 'w') as f:
-            yaml.dump(self, f)
+            f.write(inner)
```
SarahAlidoost: This works fine, but I have a suggestion to refactor these lines:

```python
import copy 

        content = copy.copy(self)
        del content.directory
        with open(target, 'w') as f:
            yaml.dump(content, f)
```

What do you think? it is also possible to use deepcopy() instead of copy().

SarahAlidoost (approved):
@sverhoeven thank you. The tests are passed. I have one suggestion to refactor some lines. Please feel free to ignore. Another issue is about `mypy` errors, see below. Shall we fix them in this PR? 

```python
mypy ewatercycle/forcing/default.py
ewatercycle/forcing/default.py:22: error: Incompatible types in assignment (expression has type "None", variable has type "str")

mypy ewatercycle/forcing/__init__.py 
ewatercycle/forcing/__init__.py:44: error: "type" has no attribute "generate"
ewatercycle/forcing/__init__.py:133: error: Argument after ** must be a mapping
```

SarahAlidoost:
@sverhoeven could you please address the merge conflicts? Thanks.

ewatercycle/forcing/default.py
sverhoeven: Thanks, that is much more readable. 

The deepcopy() is not needed as the clone only gets a string attribute deleted and the shallow copy() already takes care of copying that.

sverhoeven:
Conflicts resolved and suggestion implemented.

Thanks for reviewing

sverhoeven:
Seems the mypy errors where solved in master as running locally returns no issues.

PR #107, opened 2021-06-15
Parse dates and paths on input

Peter9192:
WIP. So far this adds a pathparser utility. 

The idea is that we use `utils.get_time` and `utils.parse_path` etc. on all inputs. 

Once we step away from dataclasses in the forcing module (in #97), we'll be able to apply those functions there. We can do the same on parameter sets and model instantiation, etc.

It requires changes to the model classes, as they currently assume the internal representation of paths and dates in the forcing objects to be strings.

### DIFF
--- a/ewatercycle/util.py
+++ b/ewatercycle/util.py
@@ -1,11 +1,10 @@
-from typing import Any, Tuple, Dict
+from datetime import datetime
+from pathlib import Path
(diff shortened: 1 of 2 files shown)


Peter9192:
Now that #143 is merged, this can be closed.

Issue #70, opened 2021-04-21
get_grdc_data should use CFG['grdc_location'] internally


sverhoeven:
At the moment I have to use

```python
observations = get_grdc_data(
    station_id=grdc_station_id,
    start_date=start_date.strftime("%Y-%m-%d"),
    end_date=end_date.strftime("%Y-%m-%d"),
   data_home=CFG["grdc_location"]
)
```
When I have `~/.config/.ewatercycle/ewatercycle.yaml` with `grdc_location: /data/grdc`.

It would be nice I if I could leave out data_home argument and internally the CFG["grdc_location"] is used.
```python
observations = get_grdc_data(
    station_id=grdc_station_id,
    start_date=start_date.strftime("%Y-%m-%d"),
    end_date=end_date.strftime("%Y-%m-%d")
)
```

PR #108, opened 2021-06-16
Implement `set`- and `get_value_at coords`

### LINKED ISSUE #53: Add `get_value_at_indices` to public interface of ewatercycle model classes.
and `set_value_at_indices` as well.

SarahAlidoost:
The new functions are also implemented for lisflood and shown in the lisflood example notebook in this PR. To check if new functions return correct values in the example notebook, you can use [this notebook](https://github.com/eWaterCycle/era5-comparison/blob/master/lisflood/02_lisflood_testrun.ipynb) as a reference: the model indices and their coordinates for some of the catchments are shown in cell No. 9.

Closes #53

### DIFF
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -3,7 +3,6 @@ from os import PathLike
 from typing import Tuple, Iterable, Any
 
 import numpy as np
-import warnings
 import xarray as xr
 from basic_modeling_interface import Bmi
 
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -123,6 +123,31 @@ class AbstractModel(metaclass=ABCMeta):
 
         """
 
+    def get_value_at_coord(self, name, lat: float, lon: float) -> np.ndarray:
+        """Get a copy of values of the given variable at lat/lon coordinates.
+
+        Args:
+            name: Name of variable
+            lat: Latitudinal value
+            lon: Longitudinal value
+
+        """
+        index = self.coord_to_index(lat,lon)
+        return self.get_value_at_indices(name, index)
+
+    def set_value_at_coord(self, name, lat: float, lon: float, value: np.ndarray) -> None:
+        """Specify a new value for a model variable at at lat/lon coordinates.
+
+        Args:
+            name: Name of variable
+            lat: Latitudinal value
+            lon: Longitudinal value
+            value: The new value for the specified variable.
+
+        """
+        index = self.coord_to_index(lat,lon)
+        return self.set_value_at_indices(name, index, value)
+
     @property
     @abstractmethod
     def parameters(self) -> Iterable[Tuple[str, Any]]:
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -111,7 +111,7 @@ class AbstractModel(metaclass=ABCMeta):
         Args: index: The index into lat/lon values
 
         """
-        raise NotImplementedError("No generic indices_to_coords method available.")
+        raise NotImplementedError("Method to convert from indices to model coordinates not implemented for this model.")
 
     @abstractmethod
     def get_value_as_xarray(self, name: str) -> xr.DataArray:
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -67,7 +67,7 @@ class AbstractModel(metaclass=ABCMeta):
             lon: Longitudinal value
 
         """
-        indices = self._coords_to_indices(name, lat, lon)
+        indices, _, _ = self._coords_to_indices(name, lat, lon)
         indices = np.array(indices)
         return self.bmi.get_value_at_indices(name, indices)
 
@@ -91,11 +91,11 @@ class AbstractModel(metaclass=ABCMeta):
             value: The new value for the specified variable.
 
         """
-        indices = self._coords_to_indices(name, lat, lon)
+        indices, _, _ = self._coords_to_indices(name, lat, lon)
         indices = np.array(indices)
         self.bmi.set_value_at_indices(name, indices, values)
 
-    def _coords_to_indices(self, name: str, lat: Iterable[float], lon: Iterable[float]) -> Iterable[int]:
+    def _coords_to_indices(self, name: str, lat: Iterable[float], lon: Iterable[float]) -> Tuple[Iterable[int], Iterable[float], Iterable[float]]:
         """Converts lat/lon values to index.
 
         Args:
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -3,6 +3,7 @@ from os import PathLike
 from typing import Tuple, Iterable, Any
 
 import numpy as np
+import warnings
 import xarray as xr
 from basic_modeling_interface import Bmi
 
@@ -107,6 +108,9 @@ class AbstractModel(metaclass=ABCMeta):
             index: The index into the variable array
 
         """
+
+    def _conversion_feedback(self, coord_user, coord_converted):
+        warnings.warn(f"Your coordinate {coord_user} matches model coordinate {coord_converted}.")
 
     @abstractmethod
     def get_value_as_xarray(self, name: str) -> xr.DataArray:
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -59,7 +59,7 @@ class AbstractModel(metaclass=ABCMeta):
         """
         return self.bmi.get_value(name)
 
-    def get_value_at_indices(self, name: str, indices: np.ndarray) -> np.ndarray:
+    def get_value_at_indices(self, name: str, indices: Iterable[int]) -> np.ndarray:
         """Get a copy of values of the given variable at particular indices.
 
         Args:
@@ -67,9 +67,10 @@ class AbstractModel(metaclass=ABCMeta):
             indices: The indices into the variable array
 
         """
+        indices = np.array(indices)
         return self.bmi.get_value_at_indices(name, indices)
 
-    def get_value_at_coords(self, name, lat: np.ndarray, lon: np.ndarray) -> np.ndarray:
+    def get_value_at_coords(self, name, lat: Iterable[float], lon: Iterable[float]) -> np.ndarray:
         """Get a copy of values of the given variable at lat/lon coordinates.
 
         Args:
@@ -91,7 +92,7 @@ class AbstractModel(metaclass=ABCMeta):
         """
         self.bmi.set_value(name, value)
 
-    def set_value_at_indices(self, name: str, indices: np.ndarray, value: np.ndarray) -> None:
+    def set_value_at_indices(self, name: str, indices: Iterable[int], value: np.ndarray) -> None:
         """Specify a new value for a model variable at particular indices.
 
         Args:
@@ -100,9 +101,10 @@ class AbstractModel(metaclass=ABCMeta):
             value: The new value for the specified variable.
 
         """
+        indices = np.array(indices)
         self.bmi.set_value_at_indices(name, indices, value)
 
-    def set_value_at_coords(self, name: str, lat: np.ndarray, lon: np.ndarray, values: np.ndarray) -> None:
+    def set_value_at_coords(self, name: str, lat: Iterable[float], lon: Iterable[float], values: np.ndarray) -> None:
         """Specify a new value for a model variable at at lat/lon coordinates.
 
         Args:
@@ -115,12 +117,11 @@ class AbstractModel(metaclass=ABCMeta):
         indices = self.coords_to_indices(name, lat, lon)
         self.set_value_at_indices(name, indices, values)
 
-    # TODO Fix this function
     def conversion_feedback(self, coord_user, coord_converted):
-        warnings.warn(f"Your coordinates {coord_user} matches model coordinates {coord_converted}.")
+        print(f"Your coordinates {coord_user} match these model coordinates {coord_converted}.")
 
     @abstractmethod
-    def coords_to_indices(self, name: str, lat: np.ndarray, lon: np.ndarray) -> np.ndarray:
+    def coords_to_indices(self, name: str, lat: Iterable[float], lon: Iterable[float]) -> Iterable[int]:
         """Converts lat/lon values to index.
 
         Args:
@@ -130,7 +131,7 @@ class AbstractModel(metaclass=ABCMeta):
         """
 
     @abstractmethod
-    def indices_to_coords(self, name: str, indices: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    def indices_to_coords(self, name: str, indices: Iterable[int]) -> Tuple[Iterable[float], Iterable[float]]:
         """Convert index to lat/lon values.
 
         Args: index: The index into lat/lon values
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -59,7 +59,7 @@ class AbstractModel(metaclass=ABCMeta):
         return self.bmi.get_value(name)
 
     def get_value_at_indices(self, name: str, indices: np.ndarray) -> np.ndarray:
-        """Get a copy of values of the given variable.
+        """Get a copy of values of the given variable at particular indices.
 
         Args:
             name: Name of variable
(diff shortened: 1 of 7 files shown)


ewatercycle/models/abstract.py
```diff
@@ -58,6 +58,29 @@ def get_value(self, name: str) -> np.ndarray:
         """
         return self.bmi.get_value(name)
 
+    def get_value_at_indices(self, name: str, indices: Iterable[int]) -> np.ndarray:
```
Peter9192: In principle I like having default implementations for these functions, but since we also support models like marrmot which don't have a grid, we should be careful with that. 

Should we implement a check here (`if self.bmi.grid_type == None`) or catch exceptions?

ewatercycle/models/abstract.py
```diff
@@ -68,6 +91,52 @@ def set_value(self, name: str, value: np.ndarray) -> None:
         """
         self.bmi.set_value(name, value)
 
+    def set_value_at_indices(self, name: str, indices: Iterable[int], value: np.ndarray) -> None:
+        """Specify a new value for a model variable at particular indices.
+
+        Args:
+            name: Name of variable
+            indices: The indices into the variable array
+            value: The new value for the specified variable.
```
Peter9192: Should this be 1D arrays? Or can it also be multiple dimensions?

ewatercycle/models/abstract.py
```diff
@@ -68,6 +91,52 @@ def set_value(self, name: str, value: np.ndarray) -> None:
...
+            lat: Latitudinal value
+            lon: Longitudinal value
+            value: The new value for the specified variable.
+
+        """
+        indices = self.coords_to_indices(name, lat, lon)
+        self.set_value_at_indices(name, indices, values)
+
+    def conversion_feedback(self, coord_user, coord_converted):
+        print(f"Your coordinates {coord_user} match these model coordinates {coord_converted}.")
```
Peter9192: Not sure if this needs to be a default implementation, might as well be part of the actual implementation. Also it shouldn't be a public API method. I suggest moving it to the implementation in the model-specific class.

ewatercycle/models/abstract.py
```diff
@@ -68,6 +91,52 @@ def set_value(self, name: str, value: np.ndarray) -> None:
...
+    @abstractmethod
+    def coords_to_indices(self, name: str, lat: Iterable[float], lon: Iterable[float]) -> Iterable[int]:
+        """Converts lat/lon values to index.
+
+        Args:
+            lat: Latitudinal value
+            lon: Longitudinal value
+
+        """
+
```
Peter9192:
```suggestion
        """
        return NotImplementedError("nice message")

```
and remove the `@abstractmethod`, as I think it's okay for models to not implement this method.

ewatercycle/models/abstract.py
```diff
@@ -68,6 +91,52 @@ def set_value(self, name: str, value: np.ndarray) -> None:
...
+    @abstractmethod
+    def indices_to_coords(self, name: str, indices: Iterable[int]) -> Tuple[Iterable[float], Iterable[float]]:
+        """Convert index to lat/lon values.
+
+        Args: index: The index into lat/lon values
+
+        """
```
Peter9192: Same: I think this doesn't have to be an abstractmethod

ewatercycle/models/lisflood.py
```diff
@@ -241,6 +241,43 @@ def get_value_as_xarray(self, name: str) -> xr.DataArray:
...
+        shape = self.bmi.get_grid_shape(grid_id) #(len(y), len(x))
+        x_model = self.bmi.get_grid_x(grid_id)
+        y_model = self.bmi.get_grid_y(grid_id)
+        indices = []
+        coord_converted = []
+        # in lisflood, x corresponds to lon, and y to lat.
+        # this might not be the case for other models!
+        for x, y in zip(lon, lat):
+            idx = np.abs(x_model - x).argmin()
+            idy = np.abs(y_model - y).argmin()
```
Peter9192: The absolute distance to x and y independently might not yield the nearest index. Consider using the euclidean distance instead.

Peter9192 (changes requested):
Hey @SarahAlidoost, nice work! I like how you separated the concerns around how to actually obtain the indices for a specific coordinate. My comments for now mostly focus on which methods should have defaults, and which methods should be optional. 

Might be nice to also have a `set_value_from_xarray()` method, but if this is difficult we can do it later.

ewatercycle/models/abstract.py
SarahAlidoost: As discussed, this function is removed. Also, I think the implementation in bmi takes care of non-grid models.

ewatercycle/models/abstract.py
SarahAlidoost: Currently, the [get_value](https://github.com/eWaterCycle/ewatercycle/blob/a7c8d3a2e9ff4ccfaa7f5cd1e503d3e8818c9ce5/ewatercycle/models/abstract.py#L52) returns a `np.ndarray`. I think this is used for 1D, too because `np.array` is not a type.

ewatercycle/models/abstract.py
SarahAlidoost: it is moved.

ewatercycle/models/abstract.py
SarahAlidoost: your suggestion is applied.

ewatercycle/models/abstract.py
SarahAlidoost: your suggestion is applied.

ewatercycle/models/lisflood.py
SarahAlidoost: The current function [lat_lon_to_closest_variable_indices](https://github.com/eWaterCycle/ewatercycle/blob/a7c8d3a2e9ff4ccfaa7f5cd1e503d3e8818c9ce5/ewatercycle/util.py#L43) in util.py uses absolute distance. Here, it is now changed to euclidean distance. BTW, both types of calculations are only valid for small distances because coordinates are lat/lon. Also, I added a threshold.

ewatercycle/models/abstract.py
```diff
@@ -68,6 +81,38 @@ def set_value(self, name: str, value: np.ndarray) -> None:
...
+
+    def _coords_to_indices(self, name: str, lat: Iterable[float], lon: Iterable[float]) -> Iterable[int]:
+        """Converts lat/lon values to index.
+
+        Args:
+            lat: Latitudinal value
+            lon: Longitudinal value
+
+        """
+        raise NotImplementedError("No generic coords_to_indices method available.")
```
Peter9192: Not sure if this error message would be clear to a random user. Perhaps change to something like this? Same for the reverse method.

```suggestion
        raise NotImplementedError("Method to convert from coordinates to model indices not implemented for this model.")
```

ewatercycle/models/lisflood.py
```diff
@@ -241,6 +241,53 @@ def get_value_as_xarray(self, name: str) -> xr.DataArray:
...
+        spacing_model = self.bmi.get_grid_spacing(grid_id)
+
+        # Create a grid from coordinates
+        x_vectors, y_vectors = np.meshgrid(x_model, y_model)
+
+        indices = []
+        # in lisflood, x corresponds to lon, and y to lat.
+        # this might not be the case for other models!
+        for x, y in zip(lon, lat):
+            # here we use Euclidean distance, but it is not accurate as the coordinates are in lon/lat and degrees.
```
Peter9192: Probably fine like this, but you could consider making it more accurate with this simple conversion: 

lon [km] = lon [deg] * 111 [km / deg]
lat [km] = lat [deg] * 111 [km / deg] * cos( lon [deg] )

https://stackoverflow.com/a/39540339

ewatercycle/models/lisflood.py
```diff
@@ -241,6 +241,53 @@ def get_value_as_xarray(self, name: str) -> xr.DataArray:
...
+            distance = ((x_vectors - x) ** 2 + (y_vectors - y) ** 2) ** (1/2)
+            index = distance.argmin()
+            indices.append(index)
+            idy, idx = np.unravel_index(index, shape)
+            coord_converted = (round(x_model[idx], 4), round(y_model[idy], 4)) # use 4 digits in round
+            coord_user = (round(x, 4), round(y, 4))
+            # Provide feedback
+            print(f"Your coordinates {coord_user} are closest to these model coordinates {coord_converted}.")
```
Peter9192: I think @sverhoeven preferred not to use print statements in our source code. An alternative could be to return the converted coords as a second return value.

ewatercycle/models/lisflood.py
```diff
@@ -241,6 +241,53 @@ def get_value_as_xarray(self, name: str) -> xr.DataArray:
...
+            # consider a threshold
+            if distance[idy, idx] > max(spacing_model):
+                print(f"Note: this point is outside of the model grid.")
```
Peter9192: I think this should raise an exception. But perhaps we can be a bit more permissive (say, only raise if it's more than twice the grid spacing away).

Peter9192 (commented):
Hey Sarah, 
Good progress! I have a few more comments.

Also I was thinking that the only model-specific bit is getting the correct latitude and longitude from the BMI, right? Calculating euclidean distance, finding nearest point, etc. can all be reused by wflow and pcrglob. Could we move some of that code to the utils, for example? What do you think?

ewatercycle/models/lisflood.py
SarahAlidoost: now it raises an error.

ewatercycle/models/lisflood.py
SarahAlidoost: right, it is refactored.

ewatercycle/models/lisflood.py
SarahAlidoost: oh, thanks for the links. I used a similar equation mentioned here https://en.wikipedia.org/wiki/Geographical_distance. 
Meantime, I don't know why you reversed the `lat` and `lon` in your sample codes. I assume it was by mistake.

SarahAlidoost:
> Hey Sarah,
> Good progress! I have a few more comments.
> 
> Also I was thinking that the only model-specific bit is getting the correct latitude and longitude from the BMI, right? Calculating euclidean distance, finding nearest point, etc. can all be reused by wflow and pcrglob. Could we move some of that code to the utils, for example? What do you think?

Thanks for reviewing. yes, agree, I made a new function in util. I don't think we can move more codes to util. What do you think?

ewatercycle/util.py
```diff
@@ -87,6 +86,26 @@ def lat_lon_boundingbox_to_variable_indices(model, variable, lat_min, lat_max, l
     return np.array(output)
 
 
+def find_closest_point(lonitudes: Iterable[float], latitudes: Iterable[float], longitude: float, latitude: float) -> Tuple[np.ndarray, int]:
```
Peter9192:
```suggestion
def find_closest_point(longitudes: Iterable[float], latitudes: Iterable[float], longitude: float, latitude: float) -> Tuple[np.ndarray, int]:
```

or looneytunes?

also below

Peter9192: Maybe it would be good to make a bit more distinct names for grid_longitudes and point_longitude, for example.

ewatercycle/util.py
```diff
@@ -87,6 +86,26 @@ def lat_lon_boundingbox_to_variable_indices(model, variable, lat_min, lat_max, l
...
+    """Find closest grid cell to a point based on Geographical distances.
+
+    It uses Spherical Earth projected to a plane formula:
+    https://en.wikipedia.org/wiki/Geographical_distance
```
Peter9192, line 93: Nice 👍

tests/test_util.py
```diff
@@ -23,3 +24,11 @@ def test_get_time_without_tz():
 
     assert 'not in UTC' in str(excinfo.value)
 
+
+def test_find_closest_point():
```
Peter9192, line 28: Good!

tests/test_util.py
```diff
@@ -23,3 +24,11 @@ def test_get_time_without_tz():
...
+    actual_distances, actual_index = (
+        [[118.77417243, 111.22983323],[122.9552163 , 115.67902656]],
+        1)
+    expected_distances, expected_index = find_closest_point([-99.83, -99.32], [42.25, 42.21], -99.32, 43.25)
```
Peter9192: I think actual and expected are normally swapped, as in: expected is what you expect the function will return. Actual (or: `result`) is what the function actually returned

ewatercycle/models/lisflood.py
```diff
@@ -241,6 +241,49 @@ def get_value_as_xarray(self, name: str) -> xr.DataArray:
...
+            indices.append(index)
+            idy, idx = np.unravel_index(index, shape)
+            lon_converted.append(round(x_model[idx], 4)) # use 4 digits in round
+            lat_converted.append(round(y_model[idy], 4)) # use 4 digits in round
+            # consider a threshold twice of the grid spacing
+            # and convert spacing_model to km using this approximation: 1 degree ~ 111km
+            if distance[idy, idx] > max(spacing_model) * 111 * 2:
+                raise ValueError("This point is outside of the model grid.")
+
+        return np.array(indices), np.array(lon_converted), np.array(lat_converted)
```
Peter9192: Adding some whitespace and grouping all the appends can make it more readable:

```suggestion
        for x, y in zip(lon, lat):
            distance, index = find_closest_point(x_model, y_model, x, y)            
            idy, idx = np.unravel_index(index, shape)
            
            # consider a threshold twice of the grid spacing
            # and convert spacing_model to km using this approximation: 1 degree ~ 111km
            if distance[idy, idx] > max(spacing_model) * 111 * 2:
                raise ValueError("This point is outside of the model grid.")
            
            indices.append(index)
            lon_converted.append(round(x_model[idx], 4)) # use 4 digits in round
            lat_converted.append(round(y_model[idy], 4)) # use 4 digits in round

        return np.array(indices), np.array(lon_converted), np.array(lat_converted)
```

Peter9192 (changes requested):
Almost there. Few typos and small suggestions

tests/test_util.py
SarahAlidoost: right, it is fixed now.

ewatercycle/util.py
SarahAlidoost: The arguments are refactored, and typo is fixed :-)

SarahAlidoost:
> Almost there. Few typos and small suggestions

Thanks for your comments. They are addressed. I would like to make the qulity_gate green, too. Let me see if I can add more tests, somehow.

sverhoeven:
This PR will also fix #53 and #70

SarahAlidoost:
@Peter9192  Some tests are added and the function `_indices_to_coord` is removed. Could you please have another look? thanks

tests/models/test_lisflood.py
```diff
@@ -109,3 +112,61 @@ def test_setup(self, model_with_setup, tmp_path):
...
+    def get_var_grid(self, name):
+        return 0
+
+    def get_grid_shape(self, grid_id):
+        return (14, 31) #shape returns (len(y), len(x))
+
+    def get_grid_x(self, grid_id):
+         return np.array([-124.450000000003, -124.350000000003, -124.250000000003,
+                -124.150000000003, -124.050000000003, -123.950000000003,
+                -123.850000000003, -123.750000000003, -123.650000000003,
```
Peter9192, line 149: 2 questions here:

- Why use this strange high precision ending at 000003 or 00000021?
- Since it's a mock, why not make a much smaller grid of, say, `2x3`?

Peter9192 (commented):
Hi Sarah,
Great work on adding more tests! It did spawn another question...

tests/models/test_lisflood.py
SarahAlidoost: > 2 questions here:
> 
> * Why use this strange high precision ending at 000003 or 00000021?

This format is actually returned by bmi too. This was the reason where I added a` round( , 4)` in the code.  User does not see the high precision. 

> * Since it's a mock, why not make a much smaller grid of, say, `2x3`?

Agree, but the test is using a real sample data, see class `TestLFlatlonUseCase` in the test.

tests/models/test_lisflood.py
```diff
@@ -109,3 +112,61 @@ def test_setup(self, model_with_setup, tmp_path):
...
+        def test_get_value_at_coords_faraway(self, model: Lisflood):
+            model.bmi = MockedBmi()
+            with pytest.raises(ValueError) as excinfo:
+                    model.get_value_at_coords('Discharge', lon=[0.0], lat=[0.0])
+            msg = str(excinfo.value)
+            assert "This point is outside of the model grid." in msg
+
+
+class MockedBmi(Bmi):
+    def get_var_grid(self, name):
```
Peter9192:
```suggestion
class MockedBmi(Bmi):
    """Mimic a real use case with realistic shape and abitrary high precision."""
    def get_var_grid(self, name):
```

Peter9192 (approved):

PR #118, opened 2021-06-17
Add parameterset catalogue

Peter9192:
Here's an initial idea on how we could support a parameterset catalogue.

Adds
* `ewatercycle.parameter_sets.available_parameter_sets()` + `ewatercycle.parameter_sets.get_parameter_set()` as shown in api notebook
* `ewatercycle.parameter_sets.download_example_parameter_sets()` to download example parameter sets.
* System setup notebook see #120 

Changes:
*  Moved MaskMap from parameter set to `Lisflood.setup()` see #121
* In notebooks replaced SubversionCopier for parameter sets with get_parameter_set
* Applied PyCharm formatting or black to touched files
* Renamed config file in user dir from `~/.config/.ewatercycle/ewatercycle.yaml` to `~/.config/ewatercycle/ewatercycle.yaml`
* AbstractModel now uses generics so model subclasses can have their own forcing class

Removed:
* Models no longer have their own parameter set class, they all use same one

Not part of this PR:
* Remove pararmetersetdb, see #131


ewatercycle/config/_validators.py
```diff
@@ -115,36 +124,30 @@ def validate_path(value, allow_none=False):
...
+
+validate_dict_parameterset = _make_type_validator(dict, allow_none=True)
 
 _validators = {
-    'esmvaltool_config': validate_path_or_none,
-    'grdc_location': validate_path_or_none,
-    'container_engine': validate_string_or_none,
-    'singularity_dir': validate_path_or_none,
-    'output_dir': validate_path_or_none,
-    'ewatercycle_config': validate_path_or_none,
```
Peter9192, line 137: This is a breaking change. There's a reload system for the CFG object that uses this. Do we want to get rid of that?

sverhoeven: Oeps, we should write ewatercycle.yaml without `ewatercycle_config` key.

Peter9192:
Nice progress! @sverhoeven

sverhoeven:
@Peter9192 I think the PR is ready for review. What do you think?

tests/models/test_lisflood.py
```diff
@@ -2,69 +2,54 @@
...
-            root / 'maps' / 'masksmall.map',
-            mask_dir / 'model_mask',
-        )
-        return LisfloodParameterSet(
-            directory=root,
-            MaskMap=mask_dir / 'model_mask',
-            config=root / 'settings_lat_lon-Run.xml',
-        )
+    def parameterset(self, mocked_config):
+        example_parameter_set = list(example_parameter_sets())[0]
```
Peter9192: Don't really like the need for this `list(...)` wrapper. Is there a neat way to solve that in the code?

ewatercycle/models/lisflood.py
```diff
@@ -238,26 +244,27 @@ def _coords_to_indices(self, name: str, lat: Iterable[float], lon: Iterable[floa
...
     @property
     def parameters(self) -> Iterable[Tuple[str, Any]]:
         """List the parameters for this model."""
-        #TODO fix issue #60
+        # TODO fix issue #60
         parameters = [
             ('IrrigationEfficiency', self._get_textvar_value('IrrigationEfficiency')),
             ('PathRoot', str(self.parameter_set.directory)),
-            ('MaskMap', str(self.parameter_set.MaskMap.parent)),
+            ('MaskMap', self._get_textvar_value('MaskMap')),
```
Peter9192, line 265: So this `_get_textvar_value` method raises an error if `MaskMap` is not set? Is that what we want here?

tests/models/test_lisflood.py
```diff
@@ -134,6 +118,62 @@ def test_get_value_at_coords_faraway(self, model: Lisflood):
...
+            if m.bmi:
+                # Clean up container
+                del m.bmi
+
+        @pytest.fixture
+        def model_with_setup(self, tmp_path, mocked_config, model: Lisflood):
+            with patch.object(BmiClientSingularity, '__init__', return_value=None) as mocked_constructor, patch(
+                'time.strftime', return_value='42'):
+                config_file, config_dir = model.setup(
+                    MaskMap=tmp_path / 'custommask/mask.map'
```
Peter9192: What happens if this `MaskMap` argument is omitted (and it is not part of the `parameter_set`?

ewatercycle/models/abstract.py
```diff
@@ -150,3 +168,20 @@ def time_step(self) -> float:
...
+        if not self.parameter_set:
+            # Nothing to check
+            return
+        model_name = self.__class__.__name__.lower()
+        if model_name != self.parameter_set.target_model:
+            raise ValueError(f'Parameter set has wrong target model, '
+                             f'expected {model_name} got {self.parameter_set.target_model}')
+        if self.parameter_set.supported_model_versions == set():
+            logger.warning(f'Model expects parameter set to support version {self.version}, '
+                           f'but parameter set supports any version')
```
Peter9192: If the parameter set supports ("should work with"?) any version, that should be just fine, right? Why issue a warning?

ewatercycle/models/marrmot.py
```diff
@@ -262,12 +261,13 @@ def parameters(self) -> Iterable[Tuple[str, Any]]:
...
-            'gamma_distribution_phi_parameter')
+              'threshold_flow_generation_evap_change',
+              'leakage_saturated_zone_flow_coefficient',
+              'zero_deficit_base_flow_speed',
+              'baseflow_coefficient',
+              'gamma_distribution_chi_parameter',
+              'gamma_distribution_phi_parameter')
+
 
 class MarrmotM14(AbstractModel):
```
Peter9192: Should this one also be associated with MarrmotForcing?

tests/models/test_abstract.py
```diff
@@ -161,25 +178,80 @@ def test_time_step(bmi, model: MockedModel):
...
+
+    def test_any_version(self, caplog, setup_config):
+        ps = ParameterSet(
+            name='justatest',
+            directory='justatest',
+            config='justatest/config.ini',
+            target_model='mockedmodel',  # == lowered class name
+            supported_model_versions=set()
+        )
+        with caplog.at_level(logging.WARNING):
```
Peter9192: caplog is a pytest fixture?

ewatercycle/models/abstract.py
```diff
@@ -1,19 +1,36 @@
...
+from ewatercycle.forcing import DefaultForcing
+from ewatercycle.parameter_sets import ParameterSet
 
-class AbstractModel(metaclass=ABCMeta):
+logger = logging.getLogger(__name__)
+
+ForcingT = TypeVar('ForcingT', bound=DefaultForcing)
+
+
+class AbstractModel(Generic[ForcingT], metaclass=ABCMeta):
```
Peter9192, line 18: Learned something new! Nicely done :-)

tests/parameter_sets/test_example.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.fixture
+def setup_config(tmp_path):
+    CFG['parameterset_dir'] = tmp_path
+    yield CFG
+    # Rollback changes made to CFG by tests
+    CFG.reload()
+
+
+class TestGoldenPath:
```
Peter9192: What is this?

Peter9192 (commented):
Did a first pass, haven't tested or looked at the notebook yet. I have a few comments so far.

ewatercycle/parameter_sets/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+    # TODO unpack archive in CFG['parameterset_dir'] subdirectory
+    # TODO print yaml snippet with target_model and config to add to ewatercycle.yaml
+    raise NotImplementedError("Auto download of parameter sets not yet supported")
+
+
+def example_parameter_sets() -> Iterable[ExampleParameterSet]:
+    """Lists example parameter sets that can be downloaded with :py:func:`~download_example_parameter_sets`.
+    """
+    # TODO how to add a new model docs should be updated with this part
+    return chain(
```
Peter9192: I'd rather have this return a dict or so. Then you could do:

```python
examples = ewatercycle.parameter_sets.example_parameter_sets()
print(examples)
> {'lisflood_fraser': ParameterSet('lisflood_fraser', ...), 'wflow_rhine_sbm_nc': ParameterSet('...'), ...}
lisflood_fraser = examples['lisflood_fraser']
lisflood_fraser.download()
... (run examples case for lisflood model)
```

ewatercycle/parameter_sets/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+def _parse_parametersets():
+    parametersets = {}
+    for name, options in CFG["parameter_sets"].items():
+        parameterset = ParameterSet(name=name, **options)
+        parametersets[name] = parameterset
+
+    return parametersets
+
+
+def available_parameter_sets(target_model: str = None) -> Iterable[str]:
```
Peter9192: `ewatercycle.parameter_sets.available_parameter_sets(target_model="lisflood")` currently returns a generator, which is not ideal for the public API (and not consistent with the example notebook). It would be nice if we can incorporate this example in one of our notebooks, and ideally it should return a list of parameterset names or a dict with them names as keys.

ewatercycle/parameter_sets/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+    """List available parameter sets on this machine.
+
+    Args:
+        target_model: Filter parameter sets on a model name
+
+    Returns: Names of available parameter sets on current machine.
+
+    """
+    all_parameter_sets = _parse_parametersets()
```
Peter9192, line 35: Can we print a nice message when there is no available parameterset? Currently it returns an empty generator...

ewatercycle/models/abstract.py
```diff
@@ -150,3 +168,20 @@ def time_step(self) -> float:
...
+        model_name = self.__class__.__name__.lower()
+        if model_name != self.parameter_set.target_model:
+            raise ValueError(f'Parameter set has wrong target model, '
+                             f'expected {model_name} got {self.parameter_set.target_model}')
+        if self.parameter_set.supported_model_versions == set():
+            logger.warning(f'Model expects parameter set to support version {self.version}, '
+                           f'but parameter set supports any version')
+        elif self.version not in self.parameter_set.supported_model_versions:
+            raise ValueError(
+                f'Parameter set is not supported with version {self.version} of model, '
```
Peter9192: `model = ewatercycle.models.Wflow('2020.1.13', parameters, forcing)` now returns an error about the parameter set not being compatible: `Parameter set is not supported with version 2020.1.13 of model, parameter set only supports {'2020.1.1'}`

I see two problems with this:

1. The first point of failure should be the model itself/the container. If the version is not in model.available_versions that should raise first. Only then should it proceed to check compatibility with parameterset.
2. I find the message confusing, since I'm not sure who is supposed to support what? Does the model support the parameterset? Perhaps it is clearer like this? 

```suggestion
                f'Parameter set is not compatible with version {self.version} of model, '
```

ewatercycle/parameter_sets/default.py
```diff
@@ -0,0 +1,65 @@
...
+            If Path is relative then relative to CFG['parameterset_dir'].
+        doi (str): Persistent identifier of parameter set. For a example a DOI for a Zenodo record.
+        target_model (str): Name of model that parameter set can work with
+        supported_model_versions (Set[str]): Set of model versions that are supported by this parameter set.
+            If not set then parameter set will be supported by all versions of model
+    """
+
+    def __init__(
+        self,
+        name: str,
```
Peter9192, line 24: In the example API notebook, `id` is used instead of `name`. Maybe change it, or support both?

```py
@property
def id(self):
    return self.name
```

Peter9192 (commented):
Tested with the notebooks. A few more comments about the functionality.

ewatercycle/models/abstract.py
Peter9192: Can we use "compatible with" instead of "x supports y"?

tests/models/test_lisflood.py
sverhoeven: I made `example_parameter_sets()` return a dict so you can use `example_parameter_sets()['lisflood_fraser']`to get it.

ewatercycle/models/lisflood.py
sverhoeven: At https://ec-jrc.github.io/lisflood-code/4_annex_input-files/ it says default is `area` for MaskMap. Added defaults for MaskMap and IrrigationEfficiency

tests/models/test_lisflood.py
sverhoeven: The model container will throw an exception during `initialize()`. I feel this is fine as the ewatercycle package should not try to solve bugs in the parameter set.

ewatercycle/models/abstract.py
sverhoeven: I wanted to give a warning to the user that parameter set is potentially untested with the selected model version. Maybe better to lower it to `logger.info`

ewatercycle/models/marrmot.py
sverhoeven: Oeps, you are right, fixed it.

tests/models/test_abstract.py
sverhoeven: Yep see https://docs.pytest.org/en/latest/reference/reference.html?highlight=caplog#std-fixture-caplog

ewatercycle/models/abstract.py
sverhoeven: Thanks

tests/parameter_sets/test_example.py
sverhoeven: I think I started with the idea of having tests for the happy code path grouped together and other tests for when things go wrong. But seems I did not followed through. Flattened tests again

ewatercycle/parameter_sets/__init__.py
sverhoeven: Yep, done dict implemented and used in test, but for notebooks I would like to download all examples. This is much less verbose.

ewatercycle/parameter_sets/__init__.py
sverhoeven: Added, still needs docs explaining how to add a pararmeter set

ewatercycle/parameter_sets/__init__.py
sverhoeven: Changed it to tuple of strings as that more closely matches to api notebook.

ewatercycle/parameter_sets/default.py
sverhoeven: `id` is a built-in. In PyCharm I get `Shadows built-in name 'id' ` warning. I think we should stick with `name`.

sverhoeven:
@Peter9192 could you have another look?

ewatercycle/models/abstract.py
sverhoeven: Added check for version against available_version, which happens before pararmeter set checks.

ewatercycle/models/lisflood.py
```diff
@@ -259,11 +261,11 @@ def parameters(self) -> Iterable[Tuple[str, Any]]:
         """List the parameters for this model."""
         assert self.parameter_set is not None
         assert self.forcing is not None
-        # TODO fix issue #60
+        # TODO fix issue #84
         parameters = [
-            ('IrrigationEfficiency', self._get_textvar_value('IrrigationEfficiency')),
+            ('IrrigationEfficiency', self._get_textvar_value('IrrigationEfficiency'), 0.75),
```
Peter9192: So here we pretend there's a default, even if it's not set in the model config? And then the model itself will fail?

ewatercycle/models/abstract.py
Peter9192:
```suggestion
        if self.parameter_set.supported_model_versions == set():
            logger.info(f'Model version {self.version} is not explicitly listed in the supported model versions of this parameter set. This can lead to compatibility issues.')
```

Peter9192 (commented):
Almost there. I made a suggestion for that warning message. Also two minor points about the notebooks:

- `download_example_parameter_sets` fails when one of the parameterset dirs already exists. Might be nice to have a 'skip_existing' and 'force_overwrite' option. Now, the CFG is also not updated, and my CFG.reload results in an old config file. 
- can we change `id` to `doi` in the example API notebook as well, and maybe execute all the commands in the cell that's about the parameter sets, instead of having fake output in comments?

ewatercycle/models/lisflood.py
sverhoeven: Yes, it will fail when area.map or area.nc file does not exist

sverhoeven: Reverted having a default in the model, will raise KeyError when MaskMap or IrrigationEfficiency can not be found in config.
An bmi model will return error when file from config does not exist

PR #127, opened 2021-06-24
Fix if statement in _generate_work_dir for marrmot

SarahAlidoost:
In this fix if the work_dir does not exist, then the setup() will create it. This is needed when running the calibration.


Peter9192 (approved):
Check! I remember we talked about this some time ago :p

PR #129, opened 2021-06-24
Destroy Bmi when on finalize or delete

### LINKED ISSUE #88: Terminate bmi container in model.finalize()
The model started the container so it should also clean it up.
See https://github.com/eWaterCycle/ewatercycle/blob/c73cb0c4d239e8e5b764e29e6f2f915b0d7b6853/ewatercycle/models/abstract.py#L43-L46

Peter9192:
Destroy Bmi when the user calls model.finalize or when the model object is deleted.

closes #88

### DIFF
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -17,7 +17,7 @@ class AbstractModel(metaclass=ABCMeta):
         self.bmi: Bmi = None  # bmi should set in setup() before calling its methods
 
     def __del__(self):
-        del self.bmi
+        self.bmi: Bmi = None
 
     @abstractmethod
     def setup(self, *args, **kwargs) -> Tuple[PathLike, PathLike]:
@@ -46,7 +46,7 @@ class AbstractModel(metaclass=ABCMeta):
     def finalize(self) -> None:
         """Perform tear-down tasks for the model."""
         self.bmi.finalize()
-        del self.bmi
+        self.bmi = None
 
     def update(self) -> None:
         """Advance model state by one time step."""
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -17,7 +17,10 @@ class AbstractModel(metaclass=ABCMeta):
         self.bmi: Bmi = None  # bmi should set in setup() before calling its methods
 
     def __del__(self):
-        self.bmi: Bmi = None
+        try:
+            del self.bmi
+        except AttributeError:
+            pass
 
     @abstractmethod
     def setup(self, *args, **kwargs) -> Tuple[PathLike, PathLike]:
@@ -46,7 +49,7 @@ class AbstractModel(metaclass=ABCMeta):
     def finalize(self) -> None:
         """Perform tear-down tasks for the model."""
         self.bmi.finalize()
-        self.bmi = None
+        del self.bmi
 
     def update(self) -> None:
         """Advance model state by one time step."""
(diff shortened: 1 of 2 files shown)


ewatercycle/models/abstract.py
```diff
@@ -16,6 +16,9 @@ class AbstractModel(metaclass=ABCMeta):
     def __init__(self):
         self.bmi: Bmi = None  # bmi should set in setup() before calling its methods
 
+    def __del__(self):
+        self.bmi: Bmi = None
```
SarahAlidoost: I amnot sure why you assign the `self.bmi` to `None` rather than deleting it like `del self.bmi`.  In the case of deleting the` self.bmi`, you can use `pytest.raises(AttributeError)` in  `test_finalize_resets_bmi` as:

```python
    with pytest.raises(AttributeError):
        model.bmi
```

ewatercycle/models/abstract.py
```diff
@@ -43,7 +46,7 @@ def initialize(self, config_file: str) -> None:
     def finalize(self) -> None:
         """Perform tear-down tasks for the model."""
         self.bmi.finalize()
-        # TODO terminate container if running?
+        self.bmi = None
```
SarahAlidoost:
```suggestion
        self.__del__()
```

SarahAlidoost (changes requested):
@Peter9192 thanks for the fix and tests. Just two comments. Sorry I don't know how `weakref` works. @sverhoeven could you please have a look? thanks.

ewatercycle/models/abstract.py
Peter9192: That wouldn't work, as `__del__()` deletes the entire `model` instance, not just the `bmi` attribute.

ewatercycle/models/abstract.py
Peter9192: Maybe here we could as well use `del self.bmi`, but testing this doesn't work with `pytest.raises`, since if the `model` instance is deleted, you have no reference to the `bmi` object anymore.

sverhoeven: By setting it to None, bmi object will have a reference count of zero and be cleaned up by Pythons garbage collector at a later point of time. While calliing `del self.bmi` will call the container termination code immediately.

Peter9192: Are you sure about that @sverhoeven ? The [docs on `__del__`](https://docs.python.org/3/reference/datamodel.html?highlight=del#object.__del__) state that

> `del x` doesn’t directly call `x.__del__()` — the former decrements the reference count for `x` by one, and the latter is only called when `x`’s reference count reaches zero.

I'm also fine with using `del` instead of setting it to None, though. Just thought it was nice to keep the attribute even if there's no bmi object currently assosciated with it.

ewatercycle/models/abstract.py
Peter9192: Okay, let me rephrase that. 
- `model.finalize()` is a BMI function to wrap up the model run, and as a side effect we want to tear down the container. As a result of this operation, `model.bmi` should be destroyed, but `model` itself will still exist.
- `model.__del__()` is called before the *model instance* is finalized, so e.g. when a user calls `del model` and there's no other remaining references to `model`, so it will be garbage collected. In that case, we want to explicitly ensure that the container is destroyed as well.

Although you could technically call `self.__del__()` from within `model.finalize()`, I think this is not what the `_del__()` method is intended for, and I'd rather not do that.

SarahAlidoost (approved):
Thanks @Peter9192 :+1:

sverhoeven (approved):
LGTM

PR #137, opened 2021-07-02
Return isostr type for model.time

### LINKED ISSUE #74: Return datetime type for model.time
At the moment when I want a [datetime object](https://docs.python.org/3/library/datetime.html#datetime-objects) from a model I have to do

```python
from datetime import datetime
from cftime import num2date
current_time: datetime = num2date(
    model.time,
    model.get_time_units(),
    only_use_cftime_datetimes=False
)
```

It would be nice if ewatercycle package would do the num2date call for me. 

For example by adding a property `time_as_datetime`:

```
from datetime import datetime
current_time: datetime = model.time_as_datetime
```

SarahAlidoost:
closes #74

### DIFF
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -1,6 +1,7 @@
 from abc import ABCMeta, abstractmethod
 from os import PathLike
 from typing import Tuple, Iterable, Any
+from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


sverhoeven (approved):
Looks good, just need to fix merge conflicts

PR #141, opened 2021-07-05
Setup returns strings

### LINKED ISSUE #52: `model.setup` return type for `cfg_file` should match `model.initialize` input type
According to pymt, you should be able to do

```Python
cfg_file, cfg_dir = model.setup(...)
model.initialize(cfg_file)
```

This currently doesn't work as `setup` return a Path, whereas `initialize` expects a str. I think we should make `setup` return a str.

sverhoeven:
Fixes #52

### DIFF
--- a/ewatercycle/models/abstract.py
+++ b/ewatercycle/models/abstract.py
@@ -1,6 +1,5 @@
 import logging
 from abc import ABCMeta, abstractmethod
-from os import PathLike
 from datetime import datetime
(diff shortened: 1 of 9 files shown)


SarahAlidoost (approved):
looks very good to me :laughing:

sverhoeven:
Thanks for reviewing, tests for wflow and pcrglobwb will come in other PRs.

PR #143, opened 2021-07-06
Make path consistent

### LINKED ISSUE #93: Consistent handling of paths
The ewatercycle user interface uses string formats to get path inputs. It would be helpful if we internally default to using standard `Path` representations as well. #82 implements a pathparser that converts input strings to pathlib objects. It would be nice use this throughout the package. 

It might also be nice to add some custom `__str__` methods so that the user never gets to see the internal representation.

SarahAlidoost:
closes eWaterCycle/ewatercycle#93
after merging this, close PR eWaterCycle/ewatercycle#107


ewatercycle/forcing/__init__.py
```diff
@@ -25,17 +27,15 @@ def load(directory: str) -> DefaultForcing:
...
     for forcing_cls in FORCING_CLASSES.values():
         yaml.register_class(forcing_cls)
-    content = source.read_text()
     # Set directory in yaml string to parent of yaml file
     # Because in DefaultForcing.save the directory was removed
-    abs_dir = str(source.parent.expanduser().resolve())
-    content += f'directory: {abs_dir}\n'
-    forcing_info = yaml.load(content)
+    forcing_info = yaml.load(source / FORCING_YAML)
+    forcing_info.directory = source
```
sverhoeven, line 38: Weird that this works, seems yaml load is bypassing `DefaultForcing.__init__` which requires directory argument.
The directory property in the DefaultForcing object is missing after line 37, but exists after line 38.
Nice, this implementation is a much less hacky (string manipulation).

ewatercycle/forcing/_default.py
```diff
@@ -45,11 +57,13 @@ def save(self):
...
         yaml = YAML()
         yaml.register_class(self.__class__)
-        target = Path(self.directory) / FORCING_YAML
+        target = self.directory / FORCING_YAML
         # We want to make the yaml and its parent movable,
-        # so the directory should not be included in the yaml file
+        # so the directory and shape should not be included in the yaml file
         clone = copy(self)
         del clone.directory
+        clone.shape = None
```
sverhoeven: Why forget the shape, with this a loaded forcing will never have a shape?

sverhoeven: The shape has 3 cases:
1. None -> no worries
2. Inside forcing.directory -> Whole directory is self contained and portable, I would like to keep shape, in save() we could change shape to be a relative path to the forcing.directory.
3. Outside forcing.directory -> Makes shape un-portable, in save() fine to forget it and print a info/debug message.

tests/parameter_sets/test_default_parameterset.py
```diff
@@ -37,22 +37,22 @@ def test_is_available(self, parameter_set: ParameterSet):
 
     def test_repr(self, parameter_set: ParameterSet, tmp_path):
         expected = (
-            "ParameterSet(name='justatest', "
-            f"directory=PosixPath('{tmp_path}'), "
```
sverhoeven, line 41: Thanks, changing this makes tests on Windows more likely to succeed.

tests/test_util.py
```diff
@@ -32,3 +33,37 @@ def test_find_closest_point():
...
+def test_to_absolute_path_must_exist():
+    input_path = "~/nonexistent_file.txt"
+    with pytest.raises(ValueError) as excinfo:
+        to_absolute_path(input_path, must_exist=True)
+    assert "Got non-existent path" in str(excinfo.value)
+
+
+def test_to_absolute_path_with_absolute_input_and_parent(tmp_path):
+    input_path = tmp_path / "nonexistent_file.txt"
+    parsed = to_absolute_path(input_path, parent = tmp_path)
```
sverhoeven: My editor complains about types
```suggestion
    parsed = to_absolute_path(str(input_path), parent = tmp_path)
```

tests/test_util.py
```diff
@@ -32,3 +33,37 @@ def test_find_closest_point():
...
+    expected = tmp_path / "nonexistent_file.txt"
+    assert parsed == expected
+
+
+def test_to_absolute_path_with_relative_input_and_relative_parent():
+    input_path = "nonexistent_file.txt"
+    parsed = to_absolute_path(input_path, parent = Path('.'))
+    expected =  Path.cwd() / "nonexistent_file.txt"
+    assert parsed == expected
```
sverhoeven, line 75: What happens when absolute input_path is not relative to parent?

```suggestion
    assert parsed == expected
    
def test_to_absolute_path_with_absolute_input_and_nonrelative_parent(tmp_path):
    parent = tmp_path / 'parent_dir'
    input_path = tmp_path / "nonexistent_file.txt"
    parsed = to_absolute_path(str(input_path), parent = parent)
    assert parsed == input_path
```

Test above works, but I expected an exception with `input_path (/tmp/nonexistent_file.txt) can not be made absolute to parent (/tmp/parent_dir)` message. What do you think should happen?

tests/test_util.py
```diff
@@ -32,3 +33,37 @@ def test_find_closest_point():
...
+    assert parsed == input_path
+
+
+def test_to_absolute_path_with_relative_input_and_parent(tmp_path):
+    input_path = "nonexistent_file.txt"
+    parsed = to_absolute_path(input_path, parent = tmp_path)
+    expected = tmp_path / "nonexistent_file.txt"
+    assert parsed == expected
+
+
```
sverhoeven, line 63: Can we have the similar test as `test_to_absolute_path_with_relative_input_and_relative_parent`, but without parent argument? 
```suggestion

def test_to_absolute_path_with_relative_input_and_no_parent():
    input_path = "nonexistent_file.txt"
    parsed = to_absolute_path(input_path)
    expected =  Path.cwd() / "nonexistent_file.txt"
    assert parsed == expected
    
    
```

ewatercycle/util.py
```diff
@@ -169,3 +171,16 @@ def data_files_from_recipe_output(recipe_output: RecipeOutput) -> Tuple[str, Dic
     # TODO simplify (recipe_output.location) when next esmvalcore release is made
     directory = str(data_files[0].filename.parent)
     return directory, forcing_files
+
+
+def to_absolute_path(input_path: str, parent: Path = None, must_exist: bool = False) -> Path:
+    """Parse input string as pathlib.Path object."""
```
sverhoeven: Please add full docstring explaining the arguments.

ewatercycle/util.py
```diff
@@ -169,3 +171,16 @@ def data_files_from_recipe_output(recipe_output: RecipeOutput) -> Tuple[str, Dic
...
+    pathlike = Path(input_path)
+    if parent:
+        pathlike = parent.joinpath(pathlike)
+    parsed_path = Path(pathlike).expanduser().resolve()
+    if must_exist and not parsed_path.exists():
+        raise ValueError(f"Got non-existent path {input_path}.")
+    return parsed_path
```
sverhoeven: Python has builtin exception [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) that is more specific than ValueError. 

As [Path.resolve(strict=True)](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) does exist check for you and pathlike is already a Path object, the code can be simplified to 
```suggestion
    return pathlike.expanduser().resolve(strict=must_exist)
```

ewatercycle/models/wflow.py
```diff
@@ -15,7 +15,7 @@
 from ewatercycle.models.abstract import AbstractModel
 from ewatercycle.parameter_sets import ParameterSet
 from ewatercycle.parametersetdb.config import CaseConfigParser
-from ewatercycle.util import get_time
+from ewatercycle.util import get_time, to_absolute_path
 
 
 class Wflow(AbstractModel[WflowForcing]):
```
sverhoeven, line 21: For line 59 spawned of https://github.com/eWaterCycle/ewatercycle/issues/144

ewatercycle/models/marrmot.py
```diff
@@ -38,7 +38,7 @@ def _generate_cfg_dir(cfg_dir: Path = None) -> Path:
         scratch_dir = CFG['output_dir']
         # TODO this timestamp isnot safe for parallel processing
         timestamp = datetime.datetime.now(datetime.timezone.utc).strftime("%Y%m%d_%H%M%S")
-        cfg_dir = Path(scratch_dir) / f'marrmot_{timestamp}'
+        cfg_dir = to_absolute_path(f'marrmot_{timestamp}', parent=Path(scratch_dir))
     cfg_dir.mkdir(parents=True, exist_ok=True)
     return cfg_dir
```
sverhoeven, line 44: On line 117 and 361 we could also use `to_absolute_path(cfg_dir)` instead of `Path(cfg_dir)`

ewatercycle/models/lisflood.py
```diff
@@ -94,7 +94,7 @@ def setup(self,  # type: ignore
                     str(self.forcing_dir)
```
sverhoeven, line 94: On line 87 we could also use `to_absolute_path`

ewatercycle/models/marrmot.py
sverhoeven: Even `must_exist=True` could be set.

ewatercycle/models/abstract.py
```diff
@@ -42,6 +42,16 @@ def __del__(self):
         except AttributeError:
             pass
 
+    def __str__(self):
```
sverhoeven, line 46: Nice to see `__str__` methods, Can tests be added for them?

ewatercycle/util.py
```diff
@@ -169,3 +171,16 @@ def data_files_from_recipe_output(recipe_output: RecipeOutput) -> Tuple[str, Dic
     # TODO simplify (recipe_output.location) when next esmvalcore release is made
     directory = str(data_files[0].filename.parent)
     return directory, forcing_files
+
+
+def to_absolute_path(input_path: str, parent: Path = None, must_exist: bool = False) -> Path:
```
sverhoeven, line 176: The current code does not use `must_exist=True`, can must_exist argument be dropped?

sverhoeven (changes requested):
I searched for usage of `expanduser` and `resolve` and it is now only used in `ewatercycle/util.py:to_absolute_path`, in `ewatercycle/config/_config_object.py:load_from_file` and in `ewatercycle/observation/grdc.py:get_grdc_data` . Can Config.load_from_file and the grdc be changed to also use `to_absolute_path()`?

I searched for Path constructor and mostly shows up in util.py.

This PR has a lot of smart replacements. Nice work.

See inline questions and suggestions.

ewatercycle/util.py
Peter9192: I think there are useful cases for it, but we didn't check it before. This adds an easy way to make the package more robust.

ewatercycle/models/abstract.py
```diff
@@ -42,6 +42,16 @@ def __del__(self):
...
+        """Nice formatting of model object."""
+        return "\n".join(
+            [
+                f"eWaterCycle {self.__class__.__name__}",
+                "-------------------",
+            ]
+            + [f"{k}={v!s}" for k, v in self.__dict__.items()]
```
Peter9192: make this only print the name of the model, and the name of the parameterset and forcing.
```suggestion
            + [f"Parameter set = {self.parameter_set}]
            + [f"Forcing = {self.forcing}]
            + [f"Version = {self.version}]
```

ewatercycle/util.py
sverhoeven: Ok, lets keep it, at least we have a test that excercises the argument.

tests/test_util.py
sverhoeven: Nice, test present and exception raised

ewatercycle/forcing/_default.py
sverhoeven: Now all cases are implemented and tested

ewatercycle/util.py
sverhoeven: Docstring is now present

ewatercycle/models/wflow.py
SarahAlidoost: It will be fixed in another PR.

ewatercycle/models/abstract.py
SarahAlidoost: Test are added to marrmot and wflow tests.

sverhoeven (approved):
Look goods, thanks for the additional tests.

SarahAlidoost:
> Look goods, thanks for the additional tests.

Thanks for reviewing, nice comments, and pair-programming.

PR #163, opened 2021-07-09
Update preprocessing notebook

### LINKED ISSUE #68: forcing: long esmvaltool log info in notebooks
Is there a way to summarize the main-log information or to use a scroll box in notebooks?

### LINKED ISSUE #162: Use to_absolute_path for shape in forcing for all models
For example, in this [line](https://github.com/eWaterCycle/ewatercycle/blob/0b9d2a3cd218759eae35237a1ddd28c5cea67452/ewatercycle/forcing/_lisflood.py#L80) in lisflood, `shape` should be wrapped by `to_absolute_path`. Otherwise, in `ewatercycle.forcing.generate( target_model='lisflood', ...)`, `shape` should be given as an absolute path. This should be fixed for all models.

### LINKED ISSUE #165: Wrong format start/endyear_climatology in pcrglob forcing
It should be `startyear_climatology = get_time(start_time_climatology).year `instead of `startyear_climatology= get_time(start_time_climatology)`, see this [line](https://github.com/eWaterCycle/ewatercycle/blob/33d74cb7bed8def40ef7d1cda79f79e4dcdc75cf/ewatercycle/forcing/_pcrglobwb.py#L94).

### LINKED ISSUE #176: forcing.shape is none after forcing.generate
For example, if I generate forcing for wflow with 

```python
wflow_forcing = ewatercycle.forcing.generate(
    target_model='wflow', 
    dataset='ERA5',  
    start_time='1990-01-01T00:00:00Z', 
    end_time='1990-12-31T00:00:00Z',    
    shape='./data/Rhine/Rhine.shp',
    model_specific_options={  
    'dem_file':'./wflow_rhine_sbm_nc/staticmaps/wflow_dem.map',
    }
) 
```
and then 
```python
wflow_forcing.shape
```
The output is None. It should be the path to the shapefile. see notebooks [here](https://github.com/eWaterCycle/ewatercycle/pull/163).

SarahAlidoost:
closes eWaterCycle/ewatercycle#68 closes eWaterCycle/ewatercycle#162 closes eWaterCycle/ewatercycle#165 relates eWaterCycle/ewatercycle-lisflood#11 closes eWaterCycle/ewatercycle#176

Docs preview: https://ewatercycle--163.org.readthedocs.build/en/163/examples/generate_forcing.html


ewatercycle/forcing/_lisflood.py
```diff
@@ -103,16 +105,22 @@ def generate(  # type: ignore
...
-            PrefixE0=forcing_files['e0'],
-            PrefixES0=forcing_files['es0'],
-            PrefixET0=forcing_files['et0'],
-        )
+        if run_lisvap:
+            raise NotImplementedError('Dont know how to run LISVAP.')
+        else:
+            message = (
+                f"The run_lisvap is False. Therefore, LISFLOOD forcing data 'e0', 'es0' and 'et0' are not generated. "
+                f"However, the recipe creates LISVAP input data that can be found in {directory}.")
```
Peter9192:
```suggestion
                f"Parameter `run_lisvap` is set to False. No forcing data will be generator for 'e0', 'es0' and 'et0'. "
                f"However, the recipe creates LISVAP input data that can be found in {directory}.")
```

ewatercycle/forcing/_pcrglobwb.py
```diff
@@ -66,13 +66,13 @@ def generate(  # type: ignore
...
 
-        basin = Path(shape).stem
+        basin = to_absolute_path(shape).stem
         recipe.data["diagnostics"]["diagnostic_daily"]["scripts"]["script"][
             "basin"
         ] = basin
 
         if extract_region is None:
-            extract_region = get_extents(shape)
+            extract_region = get_extents(to_absolute_path(shape))
```
Peter9192: I think I would like it even better if we started the function with: `shape = to_absolute_path(shape)` so that we have all the `to_absolute_paths` close to the user inputs, rather than everywhere throughout the code.

tests/forcing/test_lisflood.py
```diff
@@ -42,9 +42,6 @@ class MockTaskOutput:
         data_files = (
             create_netcdf('pr', tmp_path / 'lisflood_pr.nc'),
             create_netcdf('tas', tmp_path / 'lisflood_tas.nc'),
-            create_netcdf('e0', tmp_path / 'lisflood_e0.nc'),
```
Peter9192, line 45: maybe replace them with a `#TODO: add lisvap output files once implemented <issue link>`?

Peter9192 (commented):
Notebook looks good. Some additional suggestions that would make it even nicer:

- The notebooks doesn't include Marrmot or hype. Did we decide to skip those?
- Perhaps add a very short description to each model, something like: "Forcing for the PCRGlobWB is created using the ESMValTool recipe <link>. It produces two files: one for temperature and one for precipitation. You can set the start and end date, and the region. See <link to API docs> for more information.
- How difficult would it be to actually use the shapefiles in the plotting routines?
- Even more bonus: we now have plot codes for these models. Why not move them to methods in the forcing classes? So you could do `WflowForcing.plot()` :-)
- I might still make some changes to the PCRGlobWB generate function as discussed earlier today. Hope it won't cause too much trouble.

Also thanks a lot for fixing all the additional minor issues in the process!

tests/forcing/test_lisflood.py
SarahAlidoost: a TODO it is added. The lisvap input files are the recipe output.

ewatercycle/forcing/_pcrglobwb.py
SarahAlidoost: it is revised.

SarahAlidoost:
> Notebook looks good. Some additional suggestions that would make it even nicer:
> 
> * The notebooks doesn't include Marrmot or hype. Did we decide to skip those?

Marrmot generate forcing is implemented in case-1 technical notebook. I don't recall the reason why we skip the hype. :thinking: Shall we add it?

> * Perhaps add a very short description to each model, something like: "Forcing for the PCRGlobWB is created using the ESMValTool recipe . It produces two files: one for temperature and one for precipitation. You can set the start and end date, and the region. See  for more information.

it is added for each model. 

> * How difficult would it be to actually use the shapefiles in the plotting routines?
> * Even more bonus: we now have plot codes for these models. Why not move them to methods in the forcing classes? So you could do `WflowForcing.plot()` :-)

I think plotting is a new feature and not in the scope of this PR. If it is OK let's not implement it in this PR and before the release. Let's leave some work for after release. :smile:

> * I might still make some changes to the PCRGlobWB generate function as discussed earlier today. Hope it won't cause too much trouble.

Shall we make a new item and PR for that? 

> 
> Also thanks a lot for fixing all the additional minor issues in the process!

Thank you for reviewing and comments on [nbreview](https://app.reviewnb.com/eWaterCycle/ewatercycle/blob/update_preprocessing_notebook/docs%2Fexamples%2Fgenerate_forcing.ipynb%2F/discussion).

Peter9192 (approved):
Thanks Sarah!
We decided to do Hype later.

Issue #159, opened 2021-07-09
Use same singularity image format


sverhoeven:
Marrmot and lisflood use a sif file and Pcrglobwb and wflow use a `docker://` image. Both ways have pros:
* sif file can be shared between users
* `docker://` no prep needed
* `docker://` no naming convention needed for sif file

```shell
    singularity build ewatercycle-lisflood-grpc4bmi_20.10.sif docker://ewatercycle/lisflood-grpc4bmi:20.10
    singularity build ewatercycle-marrmot-grpc4bmi_2020.11.sif docker://ewatercycle/marrmot-grpc4bmi:2020.11
    singularity exec docker://ewatercycle/pcrg-grpc4bmi:setters run-bmi-server -h
    singularity exec docker://ewatercycle/wflow-grpc4bmi:2020.1.1 run-bmi-server -h
```

Peter9192:
related to #119

PR #170, opened 2021-07-12
Polishing of wflow example notebook

### LINKED ISSUE #132: Wflow example notebook requires editing generated config file
In https://github.com/eWaterCycle/ewatercycle/blob/master/docs/examples/wflow.ipynb there is a `Update the config file` chapter which asks the user to modify the config file generated by `model.setup()` before calling `model.initialize()`. While running the notebook I did not expect to have to do manual edits. I also expect users of the package will be surprised. 

Can the model class add the `[API]` section itself when missing?

As it knows which container image (version of model code) will be run and what a compatible configuration file should look like.

sverhoeven:
To make a runnable notebook this PR also fixes #132 .


ewatercycle/models/wflow.py
```diff
@@ -150,9 +162,7 @@ def _start_container(self):
                     "the singularity image was missing."
```
Peter9192, line 162: This would always raise for a new user, right? In light of #159, should we revert to using a sif file?

ewatercycle/models/wflow.py
```diff
@@ -70,6 +70,18 @@ def _setup_default_config(self):
             cfg.set("inputmapstacks", "Temperature", forcing.Temperature)
             cfg.set("run", "starttime", _iso_to_wflow(forcing.start_time))
             cfg.set("run", "endtime", _iso_to_wflow(forcing.end_time))
+        if self.version == "2020.1.1":
```
Peter9192, line 73: This block is very specific. I think it's okay for now, but eventually we should come up with something better.

tests/models/test_wflow.py
```diff
@@ -132,8 +176,7 @@ def test_get_value_as_coords(initialized_model, caplog):
         result = model.get_value_at_coords("discharge", lon=[5.2], lat=[46.8])
 
     msg = (
-        "Requested point was lon: 5.2, lat: 46.8;"
-        " closest grid point is 5.00, 47.00."
+        "Requested point was lon: 5.2, lat: 46.8;" " closest grid point is 5.00, 47.00."
```
Peter9192:
```suggestion
        "Requested point was lon: 5.2, lat: 46.8; closest grid point is 5.00, 47.00."
```

Peter9192 (changes requested):
Nice work so far! Few small comments, also see nbreview.

https://app.reviewnb.com/eWaterCycle/ewatercycle/blob/polish_wflow/docs%2Fexamples%2Fwflow.ipynb%2F/discussion

By the way: the azure item says pcrglob will become the example API notebooks.

sverhoeven:
Thanks for reviewing, I added most of your suggestions and replied to them. Can you have another look?

ewatercycle/models/wflow.py
Peter9192: @sverhoeven what do you think about this?

sverhoeven: Yep, having the image download within 15 seconds is very unlikely. That is why I wrote https://ewatercycle.readthedocs.io/en/latest/system_setup.html#model-container-images , so with properly setup system your should not get this exception. 

I suggest we change to sif file in a separate PR that works on #159.

Peter9192: but does `singularity exec docker://ewatercycle/wflow-grpc4bmi:2020.1.1 run-bmi-server -h` make the image available for all users?

sverhoeven: Nope it doesnt, its stored in ~/.singularity/cache, that is the con described in #159

Peter9192: Okay let's move it to another PR, but I think this should still be changed.

Peter9192 (approved):

PR #188, opened 2021-07-19
Update API notebook

Peter9192:
Rewrite API "tutorial" notebook into a "user guide" page that is a central element of the user documentation.
Also splits off the readme from the index page. A new index page is added with content from (a.o.) the original notebook. 

Related to #48 (added documentation page but needs to be expanded still)

Docs preview: https://ewatercycle--188.org.readthedocs.build/en/188/


SarahAlidoost:
The directory `/scratch/shared/ewatercycle/user_guide/` is accessible by ewatercycle group on Cartesius.

SarahAlidoost (approved):
@Peter9192 many thanks, looks very good :+1:

PR #198, opened 2021-07-22
Replace Zenodo hardcoded description by description from GH release …

### LINKED ISSUE #197: In Zenodo record use description from release instead of .zenodo.json:decription
The Zenodo record at https://zenodo.org/record/5119390 now has one liner description instead of much better description at https://github.com/eWaterCycle/ewatercycle/releases/tag/1.0.0 . Can the description at GH release be used on Zenodo?

Also the title has a typo is a bit to short to be unique.

sverhoeven:
…+ Make Zenodo title more unique

Fixes #197

https://zenodo.org/record/5119390 has been updated to reflect changes in this PR.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -50,7 +50,6 @@
             "orcid": "0000-0002-7508-3272"
         }
     ],
-    "description": "Python package to generate forcing, run hydrological models, compare with observations and visualize results",
     "keywords": [
         "hydrology",
         "bmi"
@@ -58,5 +57,5 @@
     "license": {
         "id": "Apache-2.0"
     },
-    "title": "ewaterycle"
+    "title": "eWaterCycle Python package"
 }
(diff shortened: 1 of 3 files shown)


Peter9192:
So if there's no `description` field in the .zenodo.json file, then it will automatically fall back to the GH release info?

Peter9192:
The markdown doesn't seem to render on the Zenodo homepage..

bouweandela:
You might also want to add the ewatercycle project under `grants` and the `nlesc` under `communities`: https://developers.zenodo.org/#representation

sverhoeven:
Rendering fixed.

sverhoeven:
> So if there's no `description` field in the .zenodo.json file, then it will automatically fall back to the GH release info?

Yep, as each Zenodo record is a specific version having GH release info makes more sense to me then a hardcoded string.

Peter9192 (approved):

PR #221, opened 2021-08-05
Add copy button to sphinx

### LINKED ISSUE #216: Add copy to clipboard button to code blocks in docs
Several code blocks in the documentation need to be copied and run in a terminal. It would be nice if I did not have to select the text and press CTRL-C, but instead press single button to fill my clipboard.

Could use https://github.com/executablebooks/sphinx-copybutton extension.

sverhoeven:
Fixes #216


Peter9192 (approved):
Works like a charm!

PR #222, opened 2021-08-05
Add all 3rd party packages that are imported to setup.py:install requ…

### LINKED ISSUE #99: Fiona not in setup
It seems `fiona` is used in `utils`, but it has not yet been added to the `setup.py`.
RTD also complains about dateutil and shapely

sverhoeven:
…ires

Fixes #99


sverhoeven:
Superseded by #227

PR #224, opened 2021-08-05
Remove CFG[esmvaltool_config]

### LINKED ISSUE #152: CFG[esmvaltool_config] unused
The configuration object has a `esmvaltool_config` key. However it is not used in any of the forcing generate() methods.

ESMValTool tries to load `~/.esmvaltool/config-user.yml` during import by default.

I dont think we need a way to tell ewatercycle to use a different configuration file. My proposal is to remove `esmvaltool_config` from `ewatercycle.CFG`.

sverhoeven:
Fixes #152

### DIFF
--- a/docs/system_setup.rst
+++ b/docs/system_setup.rst
@@ -301,7 +301,6 @@ configuration file.
 .. parsed-literal::
 
     container_engine: null
-    esmvaltool_config: None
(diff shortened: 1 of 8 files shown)


Peter9192 (approved):
Nice cleaning up :-)

Issue #190, opened 2021-07-19
Add column name argument to ewatercycle.observation.grdc.get_grdc_data


sverhoeven:
Now the pandas dataframe returned by `ewatercycle.observation.grdc.get_grdc_data()` always has `streamflow` as a column. Using it in `ewatercycle.analysis.hydrograph()` always needs renaming it to `reference` or `observation` with
```python
observations_df = observations_df.rename(columns={'streamflow': 'observation'})
```

Can the default column name be changed or an argument be added to set the column name?

SarahAlidoost:
nice idea. I think that the second option (an argument to set the name) is better. I like keeping the variable name `streamflow` as the default column name.

PR #225, opened 2021-08-05
Added column arg to get_grdc_data

sverhoeven:
Refs #190

### DIFF
--- a/ewatercycle/observation/grdc.py
+++ b/ewatercycle/observation/grdc.py
@@ -8,11 +8,13 @@ from ewatercycle.util import get_time, to_absolute_path
 
 logger = logging.getLogger(__name__)
 
+
 def get_grdc_data(station_id: str,
                   start_time: str,
                   end_time: str,
                   parameter: str = 'Q',
-                  data_home: str = None) -> Tuple[pd.core.frame.DataFrame, Dict[str, Union[str, int, float]]]:
+                  data_home: str = None,
+                  column: str ='streamflow') -> Tuple[pd.core.frame.DataFrame, Dict[str, Union[str, int, float]]]:
     """Get river discharge data from Global Runoff Data Centre (GRDC).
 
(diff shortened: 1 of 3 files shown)


ewatercycle/observation/grdc.py
```diff
@@ -8,11 +8,13 @@
...
 logger = logging.getLogger(__name__)
 
+
 def get_grdc_data(station_id: str,
                   start_time: str,
                   end_time: str,
                   parameter: str = 'Q',
-                  data_home: str = None) -> Tuple[pd.core.frame.DataFrame, Dict[str, Union[str, int, float]]]:
+                  data_home: str = None,
+                  column: str ='streamflow') -> Tuple[pd.core.frame.DataFrame, Dict[str, Union[str, int, float]]]:
```
Peter9192:
```suggestion
                  column: str = 'streamflow') -> Tuple[pd.core.frame.DataFrame, Dict[str, Union[str, int, float]]]:
```

Peter9192: And the line seems very long..

ewatercycle/observation/grdc.py
```diff
@@ -32,6 +34,7 @@ def get_grdc_data(station_id: str,
         data_home : optional. The directory where the daily grdc data is
             located. If left out will use the grdc_location in the eWaterCycle
             configuration file.
+        column: optional. Name of column in dataframe of streamflow.
```
Peter9192:
```suggestion
        column: optional. Name of column in dataframe. Default: "streamflow".
```

Peter9192 (changes requested):
Small formatting issue. Should we also update the examples? Perhaps at least this one: https://ewatercycle.readthedocs.io/en/latest/user_guide.html#Observations

sverhoeven:
Thanks for the review, I updated the user guide nb.

Peter9192 (approved):

PR #227, opened 2021-08-05
Use static metadata for package + 3rd party packages as deps + bump2version

### LINKED ISSUE #29: Include all imported 3rd party packages in setup.py:install_requires
It is best practice to include all 3rd party packages that are imported in ewatercycle/**.py  files to be included in the install_requires list in the setup.py

However dask, scipy packages are missing in setup.py. Can they be added?

### LINKED ISSUE #99: Fiona not in setup

### LINKED ISSUE #153: Use bump2version to bump versions
Now we are parsing/importing `ewatercycle.version` which gives problems at readthedocs.
We could switch to https://pypi.org/project/bump2version/ to handle the versions in the different files are bumped together.

### LINKED ISSUE #196: Missing pandoc in building documentation locally
After removing `~/.config/ewatercycle/ewatercycle.yaml `, I run `cd docs` and `make html`. it returns this error:
```
Notebook error:
PandocMissing in examples/Irrigation.ipynb:
Pandoc wasn't found.
Please check that pandoc is installed:
https://pandoc.org/installing.html
make: *** [Makefile:20: html] Error 2
```

_Originally posted by @SarahAlidoost in https://github.com/eWaterCycle/ewatercycle/issues/177#issuecomment-883264906_

sverhoeven:
Also incorperate #222

Fixes #153 
Fixes #99 
Fixes #29 
Fixes #196


docs/conf.py
```diff
@@ -57,8 +57,8 @@
 #
 # The short X.Y version.
 # The full version, including alpha/beta/rc tags.
-# TODO update version
-release = '1.0.0'
+version = '1.0.0'
```
Peter9192, line 66: Will this now automatically be changed by bump2version?

setup.cfg
```diff
@@ -1,9 +1,89 @@
...
+    Typing :: Typed
+project_urls =
+    Bug Tracker = https://github.com/eWaterCycle/ewatercycle/issues
+    Documentation = https://ewatercycle.readthedocs.io/
+    Source Code = https://github.com/eWaterCycle/ewatercycle
+
+[options]
+zip_safe = False
+python_requires = >=3.7
+packages = find:
```
Peter9192, line 37: Is there any documentation on how this works? (`find:`)

Peter9192 (commented):
Found https://stackoverflow.com/a/60885212 helpful context in reviewing this one. I wonder if it would make sense to go a step further and move all metadata to pyproject.toml as per PEP621, as suggested in that answer.

Also I find that `pip install . -e` no longer works. Is there an alternative way to build a development version?

docs/conf.py
sverhoeven: yep, by https://github.com/eWaterCycle/ewatercycle/blob/153-version-bump/.bumpversion.cfg config.

You try iit out by doing 
```shell
bump2version major
```

setup.cfg
sverhoeven: See https://setuptools.readthedocs.io/en/latest/userguide/package_discovery.html

sverhoeven:
> Found https://stackoverflow.com/a/60885212 helpful context in reviewing this one. I wonder if it would make sense to go a step further and move all metadata to pyproject.toml as per PEP621, as suggested in that answer.
> 
> Also I find that `pip install . -e` no longer works. Is there an alternative way to build a development version?

I followed https://packaging.python.org/tutorials/packaging-projects/ , except having the code in src/. As that `mv ewatercycle src/ewatercycle` should timed correctly, just be before a release or when there are 0 PRs.

Hmm `pip install . -e` works for me using setuptools v52.0.0 . I added back the setup.py.
Can you try `pip install . -e` again?

docs/conf.py
Peter9192: Check, nice!

Peter9192 (approved):

Peter9192:
Interestingly, now when I removed setup.py I was able to run `pip install -e .`

sverhoeven:
Thanks for reviewing. Spawned #228 for later.

Issue #204, opened 2021-07-26
Complete contributers


Peter9192:
Nick should be a co-author on the [Zenodo](https://zenodo.org/record/5119390) entry for the python package.. Maybe others as well?

Peter9192:
@nielsdrost @RolfHut can you add other people to this issue that should be listed as co-authors?

sverhoeven:
Nick has been added.

Waiting for @nielsdrost or @RolfHut for more authors.

RolfHut:
Nick shows as "Giesen, Nick" without the 'van de' but I guess that is an auto-import thing from ORCID to Zenodo? His ORCID is ok. From the Delft side I think we have all contributors to the package now covered. I leave it to @nielsdrost and the rest of the team to make that call for the NLeSC side.

sverhoeven:
Similar thing happened to Ben. The `van de` is present in the CITATION.cff and .zenodo.json file. Zenodo is not using the name-particle field from CITATION.cff.

Also maybe can fix Zenodo code at https://github.com/zenodo/zenodo/blob/b446db2b68ca3dc209eefcef4b9da973db039dfd/zenodo/modules/github/schemas.py#L58

I can move the `van de` to the family name field (`van de Giesen`) as a workaround.

PR #233, opened 2021-08-09
Added Nick to author list

sverhoeven:
Refs #204


Peter9192 (approved):
Only question I have is: is the .zenodo.json still necessary since zenodo added support for citation.cff recently?

sverhoeven:
I dont know how to map the communities in zenodo to citation.cff . So for now I would like to keep both.

PR #238, opened 2021-08-10
Moved /ewatercycle to /src/ewatercycle

### LINKED ISSUE #228: Move package to src/
According to https://packaging.python.org/tutorials/packaging-projects/ the preferred way is to have your package in src/. We should do this aswell. So move `/ewatercycle` to `/src/ewatercycle` and adjust the setup.cfg.

sverhoeven:
And  adjust config files to match.

Fixes #228


setup.cfg
```diff
@@ -34,6 +34,8 @@ project_urls =
 [options]
 zip_safe = False
 python_requires = >=3.7
+package_dir =
+    = src
```
Peter9192, line 38: This looks strange.. Presumably it corresponds to `package_dir = {'': 'src'}` as explained [here](https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages)

It seems to work though..

Peter9192 (approved):
I was able to install the package from this branch and use it.

sverhoeven:
Thanks for reviewing. I agree packag_dir value looks strange.

PR #239, opened 2021-08-10
Adds black and pre-commit hooks

### LINKED ISSUE #111: Consistent formatting
I don't think we've settled on a default formatter for this repo, have we? I suggest to use black, just like in the comparison.

### LINKED ISSUE #241: nrMeasurements in grdc metadata is always NA
The line number is wrong at https://github.com/eWaterCycle/ewatercycle/blob/6a4fe28ccedd356fa7549ae8f396ae57f3d7ece7/src/ewatercycle/observation/grdc.py#L269

it should be 33

sverhoeven:
Also formats files to be compliant.

Fixes #111
Fixes #241 

Tried to add flake8, but it complains a lot and all complaints have to manually fixed.

~Do we want to be flake8 compliant and drop prospector?~ Yep we do warnings first and once fixed fatal errors.


Peter9192:
Can you try to set the max line length for flake8 to 88 (black's default)? I should have some time to fix the remaining line length issues if that's ok with you

sverhoeven:
Max length is now 88 and some checks are ignored.
From the 400+ warnings only 81 are line too long errors.

Sure you can help to fix any of them

I configured pre-commit to print the flake8 warnings, but treat is as passed. 
So we can slowly fix all warnings and once done we can enforce it.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,97 @@
...
+    rev: 1.1.0
+    hooks:
+      - id: nbqa-black
+        # Match version of black used for .py and .pynb
+        additional_dependencies: [black==21.7b0]
+      - id: nbqa-isort
+        additional_dependencies: [isort==5.9.3]
+      - id: nbqa-mypy
+        additional_dependencies: [mypy==0.910]
+      # TODO renable when erros are fixed/ignored
```
Peter9192: old?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - repo: https://github.com/PyCQA/isort
+    rev: "5.9.3"
+    hooks:
+      - id: isort
+  # TODO renable when erros are fixed/ignored
+  # - repo: https://github.com/pycqa/pylint
+  #   rev: "v2.9.6"
+  #   hooks:
+  #     - id: pylint
+  # TODO renable when erros are fixed/ignored
```
Peter9192: old?

docs/examples/data/Rhine/Rhine.prj
```diff
@@ -1 +1 @@
-GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]]
\ No newline at end of file
+GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]]
```
Peter9192, line 1: Is this file supposed to be formatted?

tests/config/test_config.py
```diff
@@ -30,104 +27,112 @@ def generate_validator_testcases(valid):
...
-             for _ in ('1.5, 2.5', [1.5, 2.5], [1.5, 2.5], (1.5, 2.5),
-                       np.array((1.5, 2.5)))),
-            'fail': ((_, ValueError) for _ in ('fail', ('a', 1), (1, 2, 3)))
+            "validator": _listify_validator(validate_float, n_items=2),
+            "success": (
+                (_, [1.5, 2.5])
+                for _ in (
+                    "1.5, 2.5",
+                    [1.5, 2.5],
+                    [1.5, 2.5],
```
Peter9192, line 36: Same test twice?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
+        additional_dependencies: [types-python-dateutil]
+  - repo: https://github.com/nbQA-dev/nbQA
+    rev: 1.1.0
+    hooks:
+      - id: nbqa-black
+        # Match version of black used for .py and .pynb
```
Peter9192:
```suggestion
        # Match version of black used for .py and .ipynb
```

setup.cfg
```diff
@@ -51,42 +52,38 @@ install_requires =
...
     recommonmark
     sphinx
     sphinx_rtd_theme
-    # ipython syntax highlighting is required in doc notebooks
-    ipython
-    # release
-    bump2version
-    build
     twine
+    types-python-dateutil
```
Peter9192, line 82: Should we add nbqa and other new tools used by pre-commit to dev dependencies?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,97 @@
...
+        additional_dependencies: [black==21.7b0]
+      - id: nbqa-isort
+        additional_dependencies: [isort==5.9.3]
+      - id: nbqa-mypy
+        additional_dependencies: [mypy==0.910]
+      # TODO renable when erros are fixed/ignored
+      - id: nbqa-flake8
```
Peter9192, line 88: Could it be that the notebooks (in docs) are not picked up by these checks?

This check passes on `pre-commit run --all-files` for me, but if I do `pip install nbqa` and then `nbqa flake8 docs/*.ipynb` I get a list of issues.

Peter9192 (changes requested):
Very nice @sverhoeven , I think this really helps to improve the package! I fixed all line lengths issues except for in the notebooks. I did have a few minor comments, could you look at those please?

.pre-commit-config.yaml
MarcoGorelli: I think the issue is
```
        args: *fa
```
, I'm not sure what it's meant to do here, but if you remove it then `pre-commit run nbqa-flake8 --all-files` will fail too

setup.cfg
sverhoeven: The pre-commit hooks are running in there own virtual env so they do no need to be added to dev deps. 

Unless you want to run them outside pre-commit, but I dont think that is need because you can also do `pre-commit run black --files src/ewatercycle/util.py` to run black on a single file.

.pre-commit-config.yaml
sverhoeven: nope, https://pypi.org/project/isort/5.9.3/ and https://pypi.org/project/mypy/0.910/ are latest versions on PyPI 

Versions are there to make sure the .py and .ipynb are using same versions of linters/formatters.

.pre-commit-config.yaml
sverhoeven: Same isort,mypi or do you mean something else then version?

docs/examples/data/Rhine/Rhine.prj
sverhoeven: It is a text file according to http://wiki.gis.com/wiki/index.php/Shapefile#Shapefile_projection_format_.28.prj.29

and can still read it with
```pycon
In [1]: from cartopy.io import shapereader

In [2]: shape = shapereader.Reader('docs/examples/data/Rhine')

In [5]: list(shape.records())
Out[5]: [<Record: <shapely.geometry.polygon.Polygon object at 0x7fdfa3388a60>, {'HYBAS_ID': 2040023010, 'NEXT_DOWN': 0, 'NEXT_SINK': 2040023010, 'MAIN_BAS': 2040023010, 'DIST_SINK': 0.0, 'DIST_MAIN': 0.0, 'SUB_AREA': 163122.5, 'UP_AREA': 163122.5, 'PFAF_ID': 2326, 'SIDE': 'M', 'LAKE': 0, 'ENDO': 0, 'COAST': 0, 'ORDER': 1, 'SORT': 64, '_Slope_mea': 2.675870994597621, '_Slope_med': 1.3819899559021, '_Slope_std': 3.960913895155292, '_Slope_min': 0.0, '_Slope_max': 42.11079788208008, '_Slope_ran': 42.11079788208008, '_Slope_var': 0, '_Slope_v_1': 15.688838884834267, '_Urban_cou': 2617913.0, '_Urban_sum': 220178.0, '_Urban_mea': 0.084104399191264, '_Altitudem': -185.0, '_Altitud_1': 3944.0, '_Altituder': 4129.0, 'Urban_Perc': 8.41, '_variety': 0}, <fields>>]
```

tests/config/test_config.py
sverhoeven: Looks like it, can we fix that in a separate issue?

.pre-commit-config.yaml
sverhoeven: The `flake8` hook is verbose by default while `nbqa-flake8` was not, I forced it to be verbose so now warnings show up.

The `*fa` is a [YAML alias of YAML anchor](https://ktomk.github.io/writing/yaml-anchor-alias-and-merge-key.html) defined in `flake8` hook with `@fa` so I do not have to repeat the list of flake8 plugins.

MarcoGorelli: I didn't know you could do that, thanks! This'll be useful in other projects

Peter9192 (approved):

.pre-commit-config.yaml
Peter9192: I meant the #TODO which was above a block that was not disabled, but I guess we'll get to it once we addressed the remaining issues.

sverhoeven:
Thanks for reviewing

MarcoGorelli:
Just FYI, [black now has a Jupyter hook](https://github.com/psf/black/pull/2357), I'd advise using that instead of `nbqa-black`

PR #250, opened 2021-09-14
Increase timeout

### LINKED ISSUE #244: Make model.setup timeout configurable
Because 15 seconds (now hardcoded) is sometimes too short

Peter9192:
closes #244

### DIFF
--- a/src/ewatercycle/models/pcrglobwb.py
+++ b/src/ewatercycle/models/pcrglobwb.py
@@ -1,3 +1,5 @@
+"""eWaterCycle wrapper around PCRGlobWB BMI."""
+
 import datetime
 import logging
 from os import PathLike
@@ -24,7 +26,6 @@ class PCRGlobWB(AbstractModel[PCRGlobWBForcing]):
     """eWaterCycle implementation of PCRGlobWB hydrological model.
 
     Args:
-
         version: pick a version from :py:attr:`~available_versions`
         parameter_set: instance of
             :py:class:`~ewatercycle.parameter_sets.default.ParameterSet`.
@@ -41,6 +42,7 @@ class PCRGlobWB(AbstractModel[PCRGlobWBForcing]):
         parameter_set: ParameterSet,
         forcing: Optional[PCRGlobWBForcing] = None,
     ):
+        """Instantiate PCRGlob model object."""
         super().__init__(version, parameter_set, forcing)
         self._set_docker_image()
         self._setup_default_config()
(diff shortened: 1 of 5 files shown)


Peter9192:
@sverhoeven I made the PR but running into problems with ansible (again..). If you have a running VM, could you try if this solves the issues with wflow-rhine `model.setup` for you?

src/ewatercycle/models/pcrglobwb.py
```diff
@@ -41,6 +42,7 @@ def __init__(
         parameter_set: ParameterSet,
         forcing: Optional[PCRGlobWBForcing] = None,
     ):
+        """Instantiate PCRGlob model object."""
```
sverhoeven: This string does not show up in sphinx html files as the docstring at 24-34 takes precedences. However in  pycharm this string is shown in popup instead of the class docstring. When I remove this line then pycharm shows class docstring again. Can you remove this docstring as it is not rendered and is less helpful then the class doc string.

src/ewatercycle/models/wflow.py
```diff
@@ -41,7 +43,7 @@ def __init__(
         parameter_set: ParameterSet,
         forcing: Optional[WflowForcing] = None,
     ):
-
+        """Instantiate Wflow model object."""
```
sverhoeven: Same as for pcrglob, please remove this doc string.

sverhoeven (changes requested):
I created a Vagrant VM, made sure rclone cache was empty and installed this branch with:
```
sudo -i
pip uninstall ewatercycle
pip install git+https://github.com/eWatercycle/ewatercycle.git@increase-timeout
```

The wflow.setup() took 33s and pcrglobwb.setup() took 42s and they both finished successfully.

Lisflood before waited forever, but now only for 300s.  running lisflood.setup() took 42s and still works. Same for Marrmot models, setup() took 28s.

Nice, I also see you added/improved some more docstrings and solved some lint warnings.

src/ewatercycle/models/pcrglobwb.py
Peter9192: Done. Also removed for marrmot and lisflood.

src/ewatercycle/models/wflow.py
Peter9192: Done. Notice that the linter complained about this, so I had to add noqa flags to silence the linter.

sverhoeven (approved):
Thanks, pycharm, sphinx and pre-commit are now happy or not worse.

PR #256, opened 2021-09-24
Replace `conda install` with `mamba install`

### LINKED ISSUE #136: Use `mamba install` instead of `conda install` in README
The `conda install` takes 10-20 minutes. This can be sped up by using https://github.com/mamba-org/mamba .

sverhoeven:
Fixes #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,8 @@ install those dependencies:
 
 ```shell
 wget https://raw.githubusercontent.com/eWaterCycle/ewatercycle/main/environment.yml
-conda install mamba -n base -c conda-forge -y  # 13s
-mamba env create --file environment.yml  # 2m16,854s
-conda env create --file environment.yml  # 4m17,335s
+conda install mamba -n base -c conda-forge -y
+mamba env create --file environment.yml
 conda activate ewatercycle
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -35,7 +35,9 @@ install those dependencies:
 
 ```shell
 wget https://raw.githubusercontent.com/eWaterCycle/ewatercycle/main/environment.yml
-conda env create --file environment.yml
+conda install mamba -n base -c conda-forge -y  # 13s
+mamba env create --file environment.yml  # 2m16,854s
+conda env create --file environment.yml  # 4m17,335s
```
Peter9192: I don't think it's clear that these are two alternatives, and you're supposed to pick one or the other. Can we make that more explicit?

Peter9192 (approved):
Verified I can install with mamba, and it is indeed quite fast. For the record: which version of mamba do you need so you don't have the nomp problem?

README.md
sverhoeven: I removed the `conda install` and timings.

sverhoeven:
> Verified I can install with mamba, and it is indeed quite fast. For the record: which version of mamba do you need so you don't have the nomp problem?

Version [0.15.0](https://github.com/mamba-org/mamba/blob/master/CHANGELOG.md#0150-july-9-2021) of mamba has fix. It was released on 9 July 2021

PR #258, opened 2021-09-29
Add py.typed for a typed package according to pep-0561

sverhoeven:
See https://www.python.org/dev/peps/pep-0561/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -83,6 +83,7 @@ dev =
 
 [options.package_data]
 * = *.yaml
+ewatercycle = py.typed
 
 [coverage:run]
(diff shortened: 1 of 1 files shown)


Peter9192:
Do I understand correctly that this is just some package metadata info that describes that the ewatercycle package is written in Python with type hints?

sverhoeven:
Yep it announces to users of ewatercycle package that the package has inline type annotations. See https://mypy.readthedocs.io/en/latest/installed_packages.html#creating-pep-561-compatible-packages for less formal description.

Peter9192 (approved):

PR #259, opened 2021-09-29
Upgrade CITATION.cff to v1.2.0

sverhoeven:
Also fixes name-particle Zenodo bug, see https://github.com/eWaterCycle/ewatercycle/issues/204#issuecomment-904545376


CITATION.cff
```diff
@@ -47,24 +47,28 @@ authors:
...
 repository-code: "https://github.com/eWaterCycle/ewatercycle"
 title: eWaterCycle Python package
-doi: 10.5281/zenodo.5119390
+identifiers:
+  - description: Latest version of software
+    type: doi
+    value: "10.5281/zenodo.5119389"
+  - description: Version 1.1.0
+    type: doi
+    value: "10.5281/zenodo.5175508"
```
Peter9192, line 73: Does this always refer to latest release? Should it be included in version bump?

sverhoeven: Having the latest version in CFF is a chicken-egg problem. You only get new DOI after you made release. So for now I just wanted to list a version.

Peter9192: That's what I thought... Okay fair enough.

Peter9192 (approved):

PR #260, opened 2021-09-29
Prep CHANGELOG for next release

sverhoeven:


Peter9192 (commented):
Agree with patch release to 1.1.2.

What about #238, #239, and #250?

sverhoeven:
#238 and #239 have been added. Timeout already there as #244 issue

Peter9192 (approved):
Looks good!

PR #276, opened 2021-11-29
Re-run the user guide notebook on snellius

SarahAlidoost:


sverhoeven (approved):
Was able to run notebook on snellius

Local changes
* adjust conda activate in run jupyter script
* location of Rhine.shp
* set max_parallel_tasks: 1 in esmvaltool config, with 3 recipe failed

PR #293, opened 2022-02-02
Automatic validation of your citation metadata

fdiblen:
Hello!

We noticed that your repository is using the Citation File Format to store citation metadata in CITATION.cff. This Pull Request automates validation of that file using the cffconvert GitHub Action. That way, it's a little bit easier to be robust against future changes to the CITATION.cff file.

BTW it's perfectly fine if you don't feel like accepting this Pull Request for whatever reason -- we just thought it might be helpful is all.

We found your repository using a partially automated workflow; if you have any questions about that, feel free to create an issue over at https://github.com/cffbots/filtering/issues/

On behalf of the cffbots team, @abelsiqueira / @fdiblen / @jspaaks

### DIFF
--- a/.github/workflows/cffconvert.yml
+++ b/.github/workflows/cffconvert.yml
@@ -1,0 +1,19 @@
+name: cffconvert
+
+on:
+  push:
+    paths:
+      - CITATION.cff
+
+jobs:
+  validate:
+    name: "validate"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out a copy of the repository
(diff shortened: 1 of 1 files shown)


Peter9192:
@fdiblen thanks for opening this PR. I had expected it to run immediately, but it seems it didn't. Any idea why?

abelsiqueira:
It did not run for two reasons:
- It is from an external branch, so it is not inside your repo yet
- It only runs when CITATION.cff is changed, by default.

PR #296, opened 2022-02-15
Add lisvap to forcing

SarahAlidoost:
Mostly based on the lisflood notebooks in era5-comparison.

Results of `generate_forcing` notebook are in /`scratch-shared/ewatercycle/recipe_lisflood_20220218_100026/work/diagnostic_daily/script/global`. For the purpose of review, I added a `Lisvap` folder in under this directory. `Lisvap` folder has  `model_mask_doring.nc` and `Doring_convex.shp`.


src/ewatercycle/forcing/_lisvap.py
```diff
@@ -0,0 +1,154 @@
...
+run_lisvap(version, config_file, parameterset, directory)
+"""
+
+import os
+import subprocess
+from pathlib import Path
+from typing import Tuple
+
+from ewatercycle import CFG
+from ..models.lisflood import XmlConfig
```
sverhoeven: This is causing circular import. Can you move XmlConfig to own python module to break circle

src/ewatercycle/forcing/_lisflood.py
```diff
@@ -112,23 +122,80 @@ def generate(  # type: ignore
...
         if run_lisvap:
-            raise NotImplementedError("Dont know how to run LISVAP.")
+            # Get lisvap specific options and make paths absolute
+            lisvap_config = str(to_absolute_path(run_lisvap["lisvap_config"]))
+            mask_map = str(to_absolute_path(run_lisvap["mask_map"]))
+            version = run_lisvap["version"]
+            parameterset_dir = str(to_absolute_path(run_lisvap["parameterset_dir"]))
+
+            # Reindex data because recipe cropped the data
+            # Also, create a sub dir for global dataset because xarray does not
```
sverhoeven: Naming the sub dir `global` is because of our usecase. Can we pick a more widely usable name like `reindexed` 
 or `uncropped`?

SarahAlidoost: Because the lisflood parameter set is called `lisflood_global-masked_01degree`. But now I think that "global" might not fit a different parameter-set in the future. So, perhaps `reindexed` is a better name because those files are the output of `reindex` function!

SarahAlidoost: Done!

sverhoeven (approved):
Looking good, lets merge it and make a release.

PR #298, opened 2022-02-22
Set number of bars in hydrograph

Peter9192:
This PR modifies the hydrograph function such that it uses up to a 100 bars by default. If the input data exceed 100 rows, it is downsampled. A new keyword argument `nbars` is added to make it a bit more flexible.


src/ewatercycle/analysis/__init__.py
```diff
@@ -18,6 +33,7 @@ def hydrograph(
     precipitation_units: str = "mm day$^{-1}$",
     figsize: Tuple[float, float] = (10, 10),
     filename: Union[os.PathLike, str] = None,
+    nbars: int = 100,
```
sverhoeven: Why is the default 100? If I do not set nbar argument I expected it to either 
1. not do any aggregation or 
2. do aggregation so each bar is 1 pixel wide (figsize[0] / dpi).

src/ewatercycle/analysis/__init__.py
```diff
@@ -2,11 +2,26 @@
 from typing import Tuple, Union
 
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
 from hydrostats import metrics
 from matplotlib.dates import DateFormatter
 
 
+def downsample(df, nrows=100, agg="mean"):
```
sverhoeven: This is a utility method that I feel should not be part of the public analysis API. Can it be renamed to `_downsample()` or moved to util.py .

If you want to keep it public then the arguments should have a type and docstring. The docstring should mention when aggregation is performed.

sverhoeven (changes requested):
Nice, this will make plots with duration of more days than pixels more usable.

I have some inline questions.

sverhoeven (approved):
LGTM

PR #301, opened 2022-03-29
Improved performance of forcing generation of LISFLOOD model

sverhoeven:
By using the [ESMValTool regrid() with target_grid dictionary](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.regrid) instead of extract_region preprocessor.

This uses much less memory (9Gb vs 120Gb) and is faster (5min vs hours).


src/ewatercycle/util.py
```diff
@@ -96,6 +96,27 @@ def get_extents(shapefile: Any, pad=0) -> Dict[str, float]:
     }
 
 
+def fit_extents_to_grid(extents, step=0.1, offset=0.05, ndigits=2) -> Dict[str, float]:
```
Peter9192, line 99: Nicely done!

Peter9192 (approved):
Awesome!

sverhoeven:
Thanks for reviewing. On to the next minor release

Issue #303, opened 2022-04-20
Sporadic FutureTimeoutErrors on model.setup


Peter9192:
Every now and again we encounter a grpc.FutureTimeoutError on calling model.setup. This happened at least for pcrglob, marrmotm01, and marrmotm14. There seem to be no clear circumstances that trigger the problem.

Especially for marrmot with the calibration we start many instances this is problematic, as it will break the calibration routine. The timeout is already set to 300 seconds, which should be plenty for marrmot. Perhaps increasing the delay (currently default =5) might help?

sverhoeven:
I think this has to do with the container not starting up quick enough. Exposing the grpc4bmi delay argument and setting to a couple of seconds should give enough time for the container to start.

sverhoeven:
The latest release (which includes PR #305) adds a delay argument which could be used to give the model more time to start.

PR #305, opened 2022-05-03
Wait for configurable seconds to let Marrmot containers to start up

sverhoeven:
Refs #303

<del>
For now I set it to 3 seconds.
I want to make the delay value configurable, but not sure how
1. Add delay arg to setup()
2. Add container_delay to ewatercycle.yaml

I prefer option 1, @Peter9192 which do you prefer?
</del>

Adds delay argument to Marrmot*.setup() with by 0 as default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Formatted as described on [https://keepachangelog.com](https://keepachangelog.co
 
 ## [Unreleased]
 
+### Added
+
+- delay argument to `ewatercycle.models.Marrmot*.setup()_ ([#303](https://github.com/eWaterCycle/ewatercycle/issues/303))
+
 ## [1.3.0] (2022-04-20)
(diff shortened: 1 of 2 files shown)


Peter9192:
> I prefer option 1, @Peter9192 which do you prefer?

I agree. Is 3 seconds a good default? I think it might be.

Peter9192:
> Is 3 seconds a good default? I think it might be.

On second thought, can we set the default option to be using whatever the default value is in grpc4bmi?

sverhoeven:
Implemented setup arg, @Peter9192 can you have another look?

Peter9192 (approved):
Can't seem to change the status of the code smell to "wontfix" or something like that. However, it is a consequence of our design, so let's keep it as is.

sverhoeven:
thanks for reviewing

PR #316, opened 2022-06-28
Extend and restructure documentation on adding a model

Peter9192:
Including instructions on adding forcing data

docs preview: https://ewatercycle--316.org.readthedocs.build/en/316/adding_models.html

### DIFF
--- a/docs/adding_models.rst
+++ b/docs/adding_models.rst
@@ -1,42 +1,197 @@
-Adding a model
-==============
+Adding models to eWaterCycle
+############################
 
-Integrating a new model into the eWaterCycle system involves the following steps:
+eWaterCycle has been designed to make it easy to use hydrological models. Using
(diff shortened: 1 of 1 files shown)


RolfHut (approved):
looks very good to me, makes it clear this is hard, but do-able.

docs/adding_models.rst
```diff
@@ -1,42 +1,199 @@
...
+information about the contributing process.
 
-Adding a new version of a model
--------------------------------
+.. _Add to platform:
+
+Make the new model available on a machine that's running eWaterCycle
+********************************************************************
+
+At this stage, anyone can see and use your inside eWaterCycle on their own
```
Peter9192:
```suggestion
At this stage, anyone can see and use your model inside eWaterCycle on their own
```

PR #319, opened 2022-08-23
MARRMoT example forcing file incompatible

sverhoeven:
Got error `Exception calling application: Octave evaluation error:\nerror: structure has no member 'delta_t_days'\nerror: called from:\n    initialize at line 66, column 28` while running /docs/examples/MarrmotM01.ipynb notebook.

The setup doc says to download form master branch of MARRMoT repo, but that file is incompatible with model in container image (https://hub.docker.com/layers/marrmot-grpc4bmi/ewatercycle/marrmot-grpc4bmi/2020.11/images/sha256-b067ab307a191d6cd97ac61d026f314cc63525c800fd24543a42ccab68af00aa?context=explore).

Using forcing file from same branch as container image was build from made notebook work again.

### DIFF
--- a/docs/system_setup.rst
+++ b/docs/system_setup.rst
@@ -407,12 +407,12 @@ Download example forcing
 To be able to run the Marrmot example notebooks you need a forcing file.
 You can use ``ewatercycle.forcing.generate()`` to make it or use an
 already prepared `forcing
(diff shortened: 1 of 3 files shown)


Peter9192 (approved):
Good find!

PR #321, opened 2022-10-05
Add padding for wflow forcing

Peter9192:
This PR fixes the forcing generation for wflow. Without padding, the conservative regridding scheme fails to fill the entire catchment. This leads to forcing data with lots of NaNs.

### DIFF
--- a/src/ewatercycle/forcing/_wflow.py
+++ b/src/ewatercycle/forcing/_wflow.py
@@ -71,7 +71,7 @@ class WflowForcing(DefaultForcing):
         script["dem_file"] = dem_file
 
         if extract_region is None:
-            extract_region = get_extents(shape)
+            extract_region = get_extents(shape, pad=3)
         recipe.data["preprocessors"]["rough_cutout"]["extract_region"] = extract_region
 
         recipe.data["diagnostics"]["wflow_daily"]["additional_datasets"] = [
--- a/src/ewatercycle/forcing/_wflow.py
+++ b/src/ewatercycle/forcing/_wflow.py
@@ -1,5 +1,4 @@
-"""Forcing related functionality for wflow"""
-from pathlib import Path
+"""Forcing related functionality for wflow."""
 from typing import Dict, Optional
 
 from esmvalcore.experimental import get_recipe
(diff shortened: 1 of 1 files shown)


Peter9192:
@SarahAlidoost would you be able to quickly review this PR? Perhaps @BSchilperoort you may also wanna have a look, or I could walk you through it. Whoever has time first...

src/ewatercycle/forcing/_wflow.py
```diff
@@ -71,7 +70,7 @@ def generate(  # type: ignore
         script["dem_file"] = dem_file
 
         if extract_region is None:
-            extract_region = get_extents(shape)
+            extract_region = get_extents(shape, pad=3)
```
BSchilperoort, line 73: Why choose `3` as the value of pad? Are `1` and `2` not enough? Why not choose for an even larger value?

BSchilperoort (approved):
Hi Peter, seeing as you only really changed a single line, I left you a question about that. I'm sure you probably have a good reason to choose that padding value, but I am still curious :)

src/ewatercycle/forcing/_wflow.py
Peter9192: Good question. I know the resolution of ERA-Interim is 0.75 degrees. I'm guessing (but am not sure) that you might need at least two grid cells for conservative regridding, so that would be 1.5 degrees. Of course, we're not sure that the edge of the shapefile exactly coincides with a grid cell center coordinate, so a bit of extra spacing would give us some leeway. I tried with 3 and 5 for the Savannah catchment. Both worked fine, so I chose the smaller one.

SarahAlidoost (approved):
@Peter9192 in [01_wflow_preprocessing.ipynb](https://github.com/eWaterCycle/era5-comparison/blob/master/wflow/01_wflow_preprocessing.ipynb), cell number 3, the extents are calculating using dem file instead of the shapefile and using padding=1.5 (I _think_ two pixels of era-interim). I assume that padding=3 in this PR is enough for the coverage. But it also adds more pixels to regrid (perhaps slower recipe run time).

Peter9192:
Good find. Indeed 3 deg might be overkill, although the computational load in this recipe comes from the regridding itself (which is in the diagnostic script). I think a few extra cells of padding won't do much harm.

Thanks for reviewing!

PR #323, opened 2022-11-01
fix for #322 marrmot forcing

### LINKED ISSUE #322: esmvaltool update breaks Marrmot forcing
I think an update in esmvaltool, or some other update somewhere, has broken the marrmot forcing routine in eWaterCycle.

line 73 in https://github.com/eWaterCycle/ewatercycle/blob/main/src/ewatercycle/forcing/_marrmot.py assumes that the first element in recipe_output contains the final resulting .mat file, but it does not anymore: now it points to the preproc directory. 

I think line 73 needs to point more explicitly to  the .mat file, without relying on the order of recipe_output to remain the constant?

See attached screenshot of the output:

Screenshot 2022-11-01 at 10 52 11

RolfHut:
closes #322

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
   #   hooks:
   #     - id: pylint
   # TODO renable when errors are fixed/ignored
-  - repo: https://gitlab.com/pycqa/flake8
+  - repo: https://github.com/pycqa/flake8
     rev: "3.9.2"
     hooks:
       - id: flake8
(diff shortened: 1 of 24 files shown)


Peter9192:
All green! I fixed some test that we'd broken in the process, and added new tests for the updated functionality. Also fixed some linting/formatting issues. 

@RolfHut would you be able to test it once more? I'd like to verify that I didn't break anything for your use case.

RolfHut:
I have run a test and this works again. See the attached two screenshots as proof.

Rolf
Screenshot 2022-11-02 at 15 25 53
Screenshot 2022-11-02 at 15 26 01

Peter9192 (approved):

PR #330, opened 2023-01-23
Include full copy of Apache 2.0 license

sverhoeven:
GH does not recognize the notice as Apache 2.0

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,16 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Apache Software License 2.0
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-Copyright (c) 2018, Netherlands eScience Center & Delft University of Technology
+   1. Definitions.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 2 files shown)


sverhoeven:
This will get us a silver badge on https://sqaaas.eosc-synergy.eu/

LICENSE
```diff
@@ -1,16 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
BSchilperoort, line 189: Would we want to enter something here instead of the placeholders?

BSchilperoort (approved):
Nice! I didn't see any issues (other than the placeholder year/name).
We will get the silver badge now 🥈 

Although it seems that for gold we'll need to set up a (flake8) linter (and fix the issues), so I guess we'll settle for silver.

LICENSE
sverhoeven: The copyright headers are optional so I think the placeholders are as well.

sverhoeven:
Thanks for reviewing.

We have flake8 warnings in [pre-commit](https://github.com/eWaterCycle/ewatercycle/blob/main/.pre-commit-config.yaml#L35), but solving them is a task we have been pushing into the future for a while now.

PR #334, opened 2023-02-21
Config with pydantic

### LINKED ISSUE #332: Replace config validation with pydantic validation
In #324 I wanted to print a deprecation message when `container_engine: singularity` was used in the config file. So I wanted to add an extra validation rule. 

Current validation code is copied from matplotlib and is not easy to extend. It would be nice to use [pydantic](https://docs.pydantic.dev/).

Something like

```python
ContainerEngineEnum(str, Enum):
  docker= 'docker'
  apptainer = 'apptainer'
  singularity = 'singularity'

class Config(BaseModel):
  container_engine:ContainerEngineEnum = ContainerEngineEnum.docker
  parameterset_dir: DirectoryPath
  parameter_sets: dict[str,ParameterSet]
  ...

  @validator('container_engine')
  def singularity_deprecated(self, value):
    if value == ContainerEngineEnum.singularity:
       warnings.warn(
            "singularity is deprecated, use apptainer instead",
            DeprecationWarning
        )

raw_config = <dict loaded from yaml string>
config = Config(**raw_config)
```

sverhoeven:
Fixes #332 

Also 
* moves tests for parametersetdb module to own subdir

TODO:
  * [x] append CFG.parameterset_dir to ParameterSet.config and ParameterSet.directory
  * [x] remove ParameterSetConfig workaround
  * [x] ~make CFG.save_to_file() less ugly~ do later, has TODO in code


src/ewatercycle/config/_config_object.py
```diff
@@ -1,73 +1,156 @@
...
+                    locs.append(loc)
+                error["loc"] = tuple(locs)
+            raise
 
     def load_from_file(self, filename: Union[os.PathLike, str]) -> None:
-        """Load user configuration from the given file."""
+        """Load user configuration from the given file.
+
+        The config is cleared and updated in-place.
+        """
```
Peter9192, line 134: Is this complex mechanism still necessary? Now that we have all default values, a simple `Configuration.from_file` would suffice, wouldn't it? No need to clear the config and overwrite stuff. Or do I miss something?

src/ewatercycle/parameter_set.py
```diff
@@ -0,0 +1,61 @@
...
+    CFG.parameterset_dir."""
+    doi: str = "N/A"
+    """Persistent identifier of parameter set. For a example a DOI
+    for a Zenodo record."""
+    target_model: str = "generic"
+    """Name of model that parameter set can work with."""
+    supported_model_versions: Set[str] = set()
+    """Set of model versions that are
+    supported by this parameter set. If not set then parameter set will be
+    supported by all versions of model"""
```
Peter9192, line 29:
```suggestion
    """Set of model versions that are compatible with this parameterset."""
```

I'd love to also rename the setting to `compatible_model_versions`, but that might break stuff.

src/ewatercycle/parameter_set.py
```diff
@@ -0,0 +1,61 @@
...
+
+    class Config:
+        extra = "forbid"
```
Peter9192, line 32: Why forbid extra fields? What if I discover next week that it would be really, really helpful to add a "shapefile" to my parameterset? With this setting, that would require a discussion and a new release.

src/ewatercycle/parameter_set.py
```diff
@@ -0,0 +1,61 @@
...
+    def __str__(self):
+        """Nice formatting of parameter set."""
+        return "\n".join(
+            [
+                "Parameter set",
+                "-------------",
+            ]
+            + [f"{k}={v!s}" for k, v in self.__dict__.items()]
+        )
+
```
Peter9192, line 43: Can we please remove this now in favour of the default pydantic printing? Especially their integration with Rich makes things look much better(https://docs.pydantic.dev/usage/rich/): 

```py
In [10]: from rich import print
In [11]: print(CFG)
Configuration(
    grdc_location=PosixPath('.'),
    container_engine='docker',
    apptainer_dir=PosixPath('.'),
    singularity_dir=None,
    output_dir=PosixPath('.'),
    parameterset_dir=PosixPath('.'),
    parameter_sets={},
    ewatercycle_config=None
)
```

src/ewatercycle/config/__init__.py
```diff
@@ -81,6 +91,6 @@
     # apptainer pull docker://ewatercycle/wflow-grpc4bmi:2020.1.1
 """
 
-from ._config_object import CFG, DEFAULT_CONFIG, SYSTEM_CONFIG, USER_HOME_CONFIG, Config
+from ._config_object import CFG, SYSTEM_CONFIG, USER_HOME_CONFIG, Configuration
 
-__all__ = ["CFG", "Config", "DEFAULT_CONFIG", "SYSTEM_CONFIG", "USER_HOME_CONFIG"]
+__all__ = ["CFG", "Configuration", "SYSTEM_CONFIG", "USER_HOME_CONFIG"]
```
Peter9192, line 96: Would it make sense to move documentation to the user guide, and reduce the config to a single-file module called `config.py`?

Peter9192 (approved):
Awesome, dot notation looks so much better! Perhaps we can do some more testing on a new demo machine? Fine to merge before that, and sync with other open PRs

Peter9192:
Just out of curiosity. Have you also considered using pydantic for the forcing classes?

src/ewatercycle/config/_config_object.py
sverhoeven: The ewatercycle.CFG is a single instance and can not be re-assigned aka `CFG = Configuration()` as on import time the pointer to CFG was set, any reassignments would not be visible to the rest of the code. We need all this overwriting stuff to make sure we change the single instance. 

We could switch to a [singleton pattern](https://python-patterns.guide/gang-of-four/singleton/#a-more-pythonic-implementation )where we can reassign it, but this changes the public API.

src/ewatercycle/config/__init__.py
sverhoeven: I moved the content of src/ewatercycle/config/_config_object.py to src/ewatercycle/config/__init__.py in c86f3b75a58566167d74b416d47e37414924bd63.  
I also added a link from docs/system_setup.rst to api docs.
To not increase the public API I prefixed private things with underscore.

Moving it to src/ewatercycle/config.py could be done once src/ewatercycle/config/_lisflood_versions.py has been moved to its own plugin/repo.

src/ewatercycle/parameter_set.py
sverhoeven: I am not sure about this suggestion.
With your suggestion how would a model developer know what the behavior is of an empty set. An empty set could also mean zero versions are supported/compatible.

sverhoeven: We could make it optional so 
* None means compatible with all versions 
* set() means no versions are compatible

src/ewatercycle/parameter_set.py
sverhoeven: The original implementation forbid extra fields so I retained that behavior. See https://github.com/eWaterCycle/ewatercycle/blob/121a812125956a119c21fe87bcce6919cea7da24/src/ewatercycle/config/__init__.py#L38-L39 

It helps prevent typos in the field names.

Adding fields to the ParameterSet should still be possible. The only thing is the new field should be optional, so you are be able to load old parameter sets.

src/ewatercycle/parameter_set.py
sverhoeven: For an end user I like that they now see `/bla` instead of  `PosixPath('/bla')`.

If we do that then the Forcing and ParameterSet and models have different __str__ outputs.  I think they should look similar. We should do them all on none of them?

src/ewatercycle/parameter_set.py
Peter9192: My concern was more about the word "supported" than the behavior for not setting it. Sorry if that was unclear. With respect to making it optional, perhaps we should reconsider this behaviour altogether. Imagine I make a new (backward compatible) release of the model. Then I'd like for my existing parameter sets to work out of the box.

src/ewatercycle/parameter_set.py
Peter9192: okay, I can live with that.

src/ewatercycle/parameter_set.py
Peter9192: Rather do all then. But I do agree that `PosixPath`s are not so nice.

sverhoeven: Made same in 990eb31

The notebooks in docs/ still use old format, but created https://github.com/eWaterCycle/ewatercycle/issues/344 to make it even nicer.

sverhoeven:
> Just out of curiosity. Have you also considered using pydantic for the forcing classes?

I tried at https://github.com/eWaterCycle/ewatercycle/pull/345 but got stuck on save/load.

Peter9192:
Merged as part of #340

PR #335, opened 2023-02-22
Adding models via plugins

Peter9192:
This PR changes the way eWaterCycle discovers models. Instead of importing them directly inside `ewatercycle.models`, it uses the entrypoints plugin mechanism to find all available models in the users envionrment, and exposes them through `ewatercycle.models`. 

Currently supported models are largely unchanged, as the ewatercycle python package itself can also advertise its own models as entrypoints. But now, it is possible to add new models by simply installing another package alongside ewatercycle. See https://github.com/eWaterCycle/ewatercycle-plugin for an example plugin.


Peter9192:
Continuing in #340 (re-applying changes here on top of all Stefan's PRs), and follow up #341 #342

PR #324, opened 2022-11-29
Apptainer support

sverhoeven:
Fixes #290 

This PR does not mark singularity_dir as deprecated, this is done in #334

PR #339, opened 2023-03-07
Upgrade to BMI 2.0

sverhoeven:
TODO
* [x] import bmipy instead of basic_modelling_interface
* [x] centralize starting container
* [x] Fix tests
* ~[x] Update docs~ No docs found using BMI<v2
* [x] Install grpc4bmi from pypy >=0.3 instead of GH branch, 
  * requires PR https://github.com/eWaterCycle/grpc4bmi/pull/124 merged and new version released.
* [x] Merge #334 and #324 into this branch.

### DIFF
--- a/src/ewatercycle/container.py
+++ b/src/ewatercycle/container.py
@@ -1,6 +1,6 @@
 """Container utilities."""
 from pathlib import Path
-from typing import Iterable, Literal, Mapping, Optional, TypedDict, Union
+from typing import Dict, Iterable, Literal, Mapping, Optional, Union
 
 from bmipy import Bmi
 from grpc import FutureTimeoutError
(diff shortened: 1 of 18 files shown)


src/ewatercycle/container.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def start_container(
+    work_dir: Union[str, Path],
+    image_engine: ImageForContainerEngines,
+    input_dirs: Optional[Iterable[str]] = None,
+    image_port=55555,
+    timeout=None,
+    delay=0,
+) -> Bmi:
```
Peter9192, line 34: Having this isolated from the models is great!
Would it be possible to separate the image selection from the actual starting of container? I'd like to be able to start a container by simply passing in an image (assuming that I know which container engine I'm using)

Peter9192 (commented):
Very nice to have the containers centralized. Haven't tested yet, but the code changes look good.

Peter9192:
Also could you add a changelog entry?

Peter9192:
Note that I've made #340 which solves merge conflicts between this branch and #324

src/ewatercycle/container.py
sverhoeven: I added start_apptainer_container() and start_docker_container()

Peter9192:
Merged as part of #340

PR #341, opened 2023-03-10
Plugin folder

Peter9192:
This PR moves model-specific code into their own folders inside `ewatercycle.plugins`. That makes them more portable.

To do:

- [x] Fix relative imports
- [x] Also use entrypoints for forcing ~and parameter sets?~ Think better on what to do with example parametersets.
- [ ] ~Move model-specific tests into plugins?~  --> create `ewatercycle.test_utils`? Beyond scope


Peter9192:
Superseded by #347

PR #345, opened 2023-03-22
Replace forcing classes with pydantic

sverhoeven:
TODO make pydantic work with load() and save() to/from yaml files. 
While keeping old forcing yaml files work.


sverhoeven:
Replaced by #346

Issue #338, opened 2023-02-27
Simplify adding models


Peter9192:
Currently, adding a model to eWaterCycle is cumbersome:

- eWaterCycle-specific model code lives inside the eWaterCycle repo
- Different aspects of models (setup, forcing data, etc.) live in different modules (ewatercycle.models, ewatercycle.forcing, ...)
- Models, parameter sets, and forcing data all have versions that are supposed to match

Our current structure is as follows:

```
- ewatercycle 
  - models
    - modelA
    - modelB
    - ...
  - forcing
    - modelA
    - modelB
    - ...
  - ...
```

Alternatively, we could use something like

```
ewatercycle
  - modelA
    - model
    - forcing
    - ...
  - modelB
    - model
    - forcing
    - ...
  - ...
```

Then, all model-specific quirks are contained within one folder/module/repo. This would make it easier to add new models. It might also make sense to move all specific model implementations out of the main eWaterCycle package and support them as plugins instead. Such a structure would facilitate that:

- A new release of the plugin has to ensure that model code, forcing, and parameter sets are compatible
- Using a specific version of a model is as simple as changing the version of the plugin
- Ownership of specific models/plugins lies with the model/plugin developer
- The main eWaterCycle package would reduce to simply defining the interface, plus perhaps one or two "defaults" (lumped and distributed)

This requires some work on the package architecture:

- [x] Enable plugins  (see #336, #335, #340)
- [x] Change package structure (#341, #347)
- [x] Make a default model (#354, #360, #359)
- [x] Make default forcing (also see #337)
- [x] Move model-specific implementations out of eWaterCycle

Peter9192:
Another feature that would be very helpful for development purposes is _enabling models without containers_. We could make a distinction between a `LocalModel` and a `ContainerizedModel`. Both would have to adhere to the eWaterCycle model interface, i.e. having a `setup` function that attaches BMI (even if this is not strictly necessary for the localmodel). As such, the ewatercycle `localmodel` is an intermediate step between a 'normal' BMI model and a containerized ewatercycle model.

Peter9192:
Different constructors.

Currently our models are initialized with a version tag. While this is good for checking compatibility between model, forcing, and parameterset, it would also be very useful to be able to start a model directly from a container URI or image filename. One way to approach this is by having multiple constructors:

```py
ContainerizedModel.from_version(version="2020.10", ...)

# One option is separate constructors for docker/apptainer
[...]

PR #347, opened 2023-04-05
Plugin folder

Peter9192:
This PR restructures the repository such that all model-specific things are grouped together in a model folder. 

refs #338

To do:

- [x] Also group tests per model instead of per type
- [x] Also use plugins for forcing and example parameter sets - ban all imports from plugin folder in rest of ewatercycle package (so it doesn't break if you extract all plugins to their own repos)


src/ewatercycle/defaults/bmi_handles.py
sverhoeven, line 1: BmiHandles is not being used. Remove or use?

Peter9192:
We completely refactored parameter sets. We created a new folder called base where we can store all the pydantic base classes. New todo:

- [ ] Also move models, forcing etc. to `/base`
- [ ] Make sure CI is green again
- [ ] Update docs

BSchilperoort:
@Peter9192 the sonarcloud analysis fails on this PR as it only checks the diff lines. Do we still want to fix some of these issues? Or just merge.

Peter9192:
> @Peter9192 the sonarcloud analysis fails on this PR as it only checks the diff lines. Do we still want to fix some of these issues? Or just merge.

Let's merge

BSchilperoort (approved):
:rocket:  :smile:

PR #348, opened 2023-05-30
Fix CI, tests, docs build of the plugin-branch

### LINKED ISSUE #349: API reference in docs broken
[The API reference of the docs on RTD](https://ewatercycle.readthedocs.io/en/latest/apidocs/ewatercycle.html) seems to be broken, at least for `latest`. On `stable` it seems fine.

Strangely, if I build the docs locally on the plugin branch, the API reference comes out fine.

BSchilperoort:
- [x] Fix CI workflows
- [x] move models, forcing etc. to /base
- [x] Fix tests
- [x] Fix docs
- [ ] ~~update docs~~ see #350 
- [x] Remove Python 3.11
- [x] Implement new github download

**Note:** the following tests are still skipped (broken or superceded):

- tests/plugins/lisflood/test_model.py::TestLFlatlonUseCase (see #352 )


src/ewatercycle/forcing.py
```diff
@@ -0,0 +1,50 @@
...
+            return self._raw_dict[key].load()
+        else:
+            return self._raw_dict[key]
+
+    def __getattr__(self, attr):
+        """Accesses the keys like attributes. E.g. sources.HypeForcing."""
+        if attr in self._raw_dict.keys():
+            return self.__getitem__(attr)
+        else:
+            return getattr(self._raw_dict, attr)
```
BSchilperoort, line 35: @Peter9192 `__getattr__` implements doing:
```py
sources.HypeForcing
```
However, the IDE/type analyzer does not know the type of `HypeForcing`. If you access the class in the following way, the type analyzer does know the type:
```py
sources["HypeForcing"]
```
It sees that this is a `Type[DefaultForcing]`, and should have the methods of DefaultForcing available (`load`, `generate`...).

This is the same in xarray with:
```py
ds["temperature"].  # gives suggestions
ds.temperature. # does not give suggestions
```

It might be nicer to only have the square brackets to access the Forcings.

Peter9192: Can't we have both and leave it up to the user (like xarray)?

Peter9192 (approved):
Pair-reviewed, nice work!

Issue #350, opened 2023-06-07
Update outdated docs


BSchilperoort:
- ~~Add docstring to DefaultForcing.generate. Remove from child classes (or add to the parent classes' docstrings).~~ https://github.com/eWaterCycle/ewatercycle/pull/393
- ~~Update user guide notebook~~ https://github.com/eWaterCycle/ewatercycle/pull/393
- Validate "Update system setup" section:
  - Download example parameter sets
- Fix examples:
  - move docs to their plugins?
  - add examples with leaky buckets
- Adding models to eWaterCycle:
  - Split into "I am writing a new model" vs "I already have my own model", see #338 
- consider moving "Migrate from HPC to Cluster (Snellius) guide" out of the main docs structure. Under e.g. "infrastructure specific instructions"

From https://github.com/eWaterCycle/ewatercycle/issues/359:
- Also include complete rewrite of adding models and using models with plugin structure (https://github.com/eWaterCycle/ewatercycle/pull/383)

BSchilperoort:
I believe all these points have been tackled by now.

Issue #104, opened 2021-06-11
Bug: wrong x and y in lat_lon_to_closest_variable_indices and lat_lon_boundingbox_to_variable_indices


SarahAlidoost:
The longitude value is obtained by **x** coordinates whereas latitude by **y**:
```python
"longitude": bmi.get_grid_x(grid),
"latitude": bmi.get_grid_y(grid),
```

This is not the case in two functions implemented in util.py:
[lat_lon_to_closest_variable_indices](https://github.com/eWaterCycle/ewatercycle/blob/1156a4ff7cdc1fc7143d0cc1137695c405581f55/ewatercycle/util.py#L43)
[lat_lon_boundingbox_to_variable_indices](https://github.com/eWaterCycle/ewatercycle/blob/1156a4ff7cdc1fc7143d0cc1137695c405581f55/ewatercycle/util.py#L69)

Peter9192:
I think we've seen this before. Could it be that this was originally coded like that for wflow (or perhaps it was pcrglob)? I think one of these models might have reversed this convention internally.

SarahAlidoost:
yes, I am exploring this now. Here the documentation of [grpc4bmi for these functions](https://grpc4bmi.readthedocs.io/en/0.2.x/api/grpc4bmi.bmi_grpc_client.html#grpc4bmi.bmi_grpc_client.BmiClient.get_grid_x) that is not clear to me what is meant by **streamwise and transverse** directions.  I see that it is reversed in the [pcrglob comparison notebook ](https://github.com/eWaterCycle/era5-comparison/blob/master/pcrglobwb/03_pcrglobwb_fullruns.ipynb), and [wflow notebook](https://github.com/eWaterCycle/era5-comparison/blob/master/wflow/03_wflow_fullruns.ipynb).  But [lisflood](https://github.com/eWaterCycle/era5-comparison/blob/master/lisflood/03_lisflood_fullruns.ipynb) uses the correct notations, y for latitude and x for longitude. Not confusing at all.

SarahAlidoost:
In `wflow bmi.py`, the function [get_grid_x](https://github.com/eWaterCycle/wflow/blob/fdb404c543c199b942a0970411301e2c3a5cad84/wflow/bmi.py#L374) is defined based on column centers.

Peter9192:
> 
> 
[...]

PR #353, opened 2023-06-13
Update outdated docs

sverhoeven:
Refs #350

TODO
* [x] docstrings
  * [x] forcings
  * [x] parameter sets
  * [x] models
* [x] user guide notebook
  * [x] until model run
  * [x] generate
  * [x] model run
* [x] example notebooks
  * [x] model specific, should be moved to docs/examples/plugins/<modelname>/*.ipynb similar to src/ewatercycle/plugins and tests/plugins
  * [x] forcing generate notebook: each model section should be move to  docs/examples/plugins/<modelname>/forcing.ipynb
  * [x] Irrigation.ipynb
* [x] ~~docs on how to add model~~ Do later in PR for #338
  * [ ] I am writing a new model
  * [ ] I already have my own model
* [x] hpc2cluster docs
* [x] show that specific model has more methods. Either by adding link to super class or showing inherited methods.


sverhoeven:
@Peter9192 @BSchilperoort In the notebook we are printing objects. 
Pydantic models are printed as a long single line, which is not very user friendly.
We could use https://python-devtools.helpmanual.io/ or https://github.com/willmcgugan/rich or https://docs.python.org/3/library/pprint.html (+.dict()).

To me rich looks nicest in a notebook, which do you prefer or leave as is?

BSchilperoort:
> @Peter9192 @BSchilperoort In the notebook we are printing objects. Pydantic models are printed as a long single line, which is not very user friendly. We could use https://python-devtools.helpmanual.io/ or https://github.com/willmcgugan/rich or https://docs.python.org/3/library/pprint.html (+.dict()).
> 
> To me rich looks nicest in a notebook, which do you prefer or leave as is?

The rich prints look good! It's also a very widely used package, so I see no reason why we wouldn't want to use it.

Peter9192:
I've also used rich before, quite like it

sverhoeven:
rich it is

sverhoeven:
Notebooks where executed in a vagrant box 
* provisioned with https://github.com/eWaterCycle/infra 
* this branch installed with sudo.
* in /etc/ewatercycle.yaml, removed output_dir and pruned parameter sets, use apptainer 
* `ln -s /usr/bin/singularity /usr/bin/apptainer` as https://github.com/eWaterCycle/infra/pull/137 not merged yet.

sverhoeven:
The wlfow `get_grid_shape()`  seems to return inverted order (x,y instead of y,x acco)  which is used to reserve numpy array. Causing things to not fit.

```
# bypass OptionalDestBmi and MemoizedBmi by calling origin.origin.
model_instance.bmi.origin.origin.get_grid_shape(1, np.empty((2,), dtype=np.float64))
array([169., 187.])
# Below works
model_instance.bmi.origin.origin.get_grid_x(1, np.empty((169,), dtype=np.float64))
model_instance.bmi.origin.origin.get_grid_y(1, np.empty((187,), dtype=np.float64))
# but OptionalDestBmi uses 187 for x and 169 for y, which fails.
```

We might needs some corrective bmi class in pyewatercycle that can swap result of `get_grid_shape()` or fix in container image, which might break other code.

BSchilperoort:
> The wlfow `get_grid_shape()` seems to return inverted order (x,y instead of y,x acco) which is used to reserve numpy array. Causing things to not fit.

Looks like the problem occurs here:
https://github.com/openstreams/wflow/blob/bdc9b7966a0caa3c826b3a1f25150186b37d683f/wflow/wflow_bmi.py#L1182-L1197

The return statement should be
```py
return [dim[5], dim[4]]
```

Seeing as the old wflow is not being developed anymore, I would vote for the fix with the least amount of work. If it can be easily patched in this repo, that might be best.

BSchilperoort:
I mentioned this issue during standup, and @SarahAlidoost pointed out that there was an old issue discussing the switched x/y coordinates: https://github.com/eWaterCycle/ewatercycle/issues/104

sverhoeven:
Not all example notebooks have been run completely. So some old output is still there.

Peter9192 (approved):
This PR contains many good changes. We will need to do several more passes through the documentation as we're updating the model and forcing API, so I'm gonna approve this as so we can use is as a good basis for further improvements in follow-up PRs

Issue #336, opened 2023-02-22
Use plugins to add models (and forcing?) to eWaterCycle


Peter9192:
Currently, to add models to eWaterCycle, one needs to modify the source code of the eWaterCycle python package (see https://ewatercycle.readthedocs.io/en/latest/adding_models.html#add-the-model-to-the-ewatercycle-python-package). This is cumbersome and can be intimidating for new users. 

An alternative would be to enable adding models as plugins. Then, you can write a plugin in a completely separate repo, and install it alongside eWaterCycle like so:

```bash
pip install ewatercycle
pip install ewatercycle-plugin  # exposes class ToyModel via an entrypoint
```

And it will automatically be discovered by eWaterCycle like so:

```py
from ewatercycle.models import ToyModel
model = ToyModel(version='latest')
```

We could make one (or a few) example repo's of eWaterCycle plugins, or a template repo with instructions on how to finalize it. 

I started a branch here to see what this would entail: https://github.com/eWaterCycle/ewatercycle/pull/335. Use https://github.com/eWaterCycle/ewatercycle-plugin to test it.

BSchilperoort:
We should add an extra pypi package `ewatercycle-collection` so you can install ewatercycle+all plugins like `ewatercycle-wflow`, `ewatercycle-leakybucket` with a single pip install.

sverhoeven:
We are now using entry points in package metadata, there are 2 other methods suggested at https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/ . 

I think entrypoints offers most flexiblity, but is harder to implement. Would be nice to have a look at those other methods to see if they could be a fit for this repo.

Peter9192:
> We are now using entry points in package metadata, there are 2 other methods suggested at https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/ .
> 
> I think entrypoints offers most flexiblity, but is harder to implement. Would be nice to have a look at those other methods to see if they could be a fit for this repo.

The options are:

1. Naming conventions. In this case, we rely on plugins adhering to a fixed naming convention, for example all plugins should be called `ewatercycle_<somename>`
2. Namespace packages. In this case, you can have a plugin named `anything` on PyPi, and as long as it follows the same structure as `ewatercycle`, for example it has a module `ewatercycle.forcing` then the content of that module is available under `ewatercycle.forcing` alongside the contents of the original `ewatercycle.forcing` module.
[...]

PR #354, opened 2023-06-13
Updated model API

Peter9192:
This PR rewrites the core eWaterCycle model class, such that

- There is a base implementation for a LocalModel (mostly for development purposes)
- There is a base implementation for a ContainerizedModel (intended for production)

Together with the new with the new plugin architecture (#336, #347), this should make it easier to add new models to eWaterCycle.

To do:

- [x] Fix typeerror 
- [x] Use the new model interface for Wflow
- [x] Use the new model interface for Marrmot
- [x] Use the new model interface for Lisflood
- [x] Use the new model interface for Hype

moved to #359:
- [ ] Update tests for abstractmodel to test the new API (once crystallized)
- [ ] Update documentation
- [ ] Remove `new_model_api.py` from repo.
- [ ] Nice to have: Extend containerimage class to check version
- [ ] Nice to have: Extend containerimage class to automatically pull docker image and convert to singularity


Peter9192:
@BSchilperoort @sverhoeven I'm gonna call it a (holi)day. Tried not to make too big of a mess.. Hope you can pick it up from here.

BSchilperoort:
I discussed the API with Peter, and we've decided to (for now) pin the container version in the model class.
Model plugin developers will have to release a new version of when a new docker version is released.

E.g.:
```py
class Wflow(ContainerizedModel):
    bmi_image: ContainerImage("ewatercycle/wflow-grpc4bmi:2020.1.3")
```
Setting the version to "latest" will have reproducibility issues & managing lists of valid versions is a lot of bookkeeping and hassle.
Of course, as `bmi_image` is a kwarg, a user can overwrite this.

Additionally, I believe this will remove the need for `ParameterSet.supported_model_versions`, as the model's parameter set is packaged with the model class in the plugin.

BSchilperoort:
@Peter9192 I have moved over Hype, and fixed the Hype tests (/hype/test_model.py). Could you have a look before I move all the rest over and fix their tests?
I also had to make some modifications to `/base/model.py`

src/ewatercycle/base/model.py
```diff
@@ -374,16 +374,18 @@ class ContainerizedModel(BaseModel):
 
     bmi_image: ContainerImage
 
+    _additional_input_dirs: list[str] = pydantic.PrivateAttr()
+
     def _make_bmi_instance(self) -> bmipy.Bmi:
-        self.additional_input_dirs = []
+        self._additional_input_dirs = []
```
Peter9192: Does that need to be stored on `self` or can we just create a local `list` on the fly?

src/ewatercycle/plugins/hype/model.py
```diff
@@ -103,80 +97,92 @@ def setup(  # type: ignore
...
+        cfg_dir.mkdir(parents=True, exist_ok=True)
 
-        # start container
-        self.bmi = start_container(
-            image_engine=_version_images[self.version],
-            work_dir=cfg_dir_as_path,
+        # copy parameter set files to cfg_dir
+        assert self.parameter_set
+        shutil.copytree(
+            src=self.parameter_set.directory, dst=cfg_dir, dirs_exist_ok=True
```
Peter9192: Is this necessary? We do it for wflow because wflow wants to store its output in the same dir as where it gets its input. But our opinion so far has been that this is bad practice, and we prefer to have a read-only input folder mounted on the model containers.

src/ewatercycle/plugins/hype/model.py
```diff
@@ -137,29 +136,33 @@ def _export_config(self) -> Path:
...
+        **kwargs,
+    ) -> Path:
+        """Make sure there is a working directory.
+
+        Args:
+            cfg_dir: If cfg dir is None or does not exist then create sub-directory
+                in CFG.output_dir
+        """
+        cfg_path = super()._make_cfg_dir(
+            cfg_dir=cfg_dir, folder_prefix="hype", **kwargs
```
Peter9192: Do we need to do this for every model now only to add the folder prefix? Would it be possible to get that from `self.__class__` or `self.__name__`?

src/ewatercycle/plugins/marrmot/model.py
```diff
@@ -77,29 +48,29 @@ class MarrmotM01(AbstractModel[MarrmotForcing]):
...
 
-    def __init__(self, version: str, forcing: MarrmotForcing):  # noqa: D107
-        super().__init__(version, forcing=forcing)
-        self._parameters = [1000.0]
-        self.store_ini = [900.0]
-        self.solver = Solver()
-        self._check_forcing(forcing)
+    _forcing_filepath: str = PrivateAttr()
+    _forcing_start_time: datetime.datetime = PrivateAttr()
+    _forcing_end_time: datetime.datetime = PrivateAttr()
```
Peter9192, line 64: I wonder if it makes sense to store all these as attributes separately on the model, or whether we could store them all in a single config object (can be a simple `dict`). Ideally we could use the models' native configuration files for that. So on creation you'd parse the template config file, update it in memory as necessary, then in `setup` save a copy with the latest changes and return that one.

Peter9192 (commented):
Hi Bart, nice work so far. I have a few comments, we can discuss them if anything is unclear.

src/ewatercycle/plugins/hype/model.py
BSchilperoort: Oh I think I messed up somewhere and copied over that code incorrectly... I now see that it indeed wasn't in the code before :facepalm:

src/ewatercycle/plugins/hype/model.py
BSchilperoort: Using `self.__class__.__name__.lower()` would probably be a nice way of retrieving the folder prefix. It would usually remove the need to define `_make_cfg_dir`

Peter9192:
> Additionally, I believe this will remove the need for `ParameterSet.supported_model_versions`, as the model's parameter set is packaged with the model class in the plugin.

This only holds for the example parameter sets. However, we envision that researchers could share their parametersets as a standalone dataset, and that you could then load that by means of a configuration file. Ideally datasets would be shared via zenodo or other repos and retrieved through their doi. So then, you'd need to add sufficient metadata about compatibility.

src/ewatercycle/base/model.py
BSchilperoort: It's actually best to keep it stored in self: this allows models to add to the additional dirs in their own class definition. I've used this for lisflood.

I have changed it so it's now initiated in the class definition, allowing appending to the list before _make_bmi... is called.

BSchilperoort:
I have now moved over all models, and attempted to standardize them a bit more. All tests except the lisflood ones (and AbstractModel one) are now passing. The lisflood ones are a bit difficult to work on because a large download takes place every time the test is run.

Working with pydantic is a bit frustrating as it doesn't seem to fit our use completely. In v1 for example, a post_init method is missing, and the `@root_validator` decorator is not a very good replacement (as it takes `cls` and not `self`). At least that's fixed in v2.

Also, pydantic seems to have an attribute "parameters" which can get a bit confusing for us/our users.

Peter9192:
> Working with pydantic is a bit frustrating as it doesn't seem to fit our use completely.

We should definetely start using v2 asap. But if that's still frustrating, we could consider using "normal" dataclasses.

src/ewatercycle/plugins/marrmot/model.py
BSchilperoort: I am not sure if it matters too much if they're separate private attributes or items of a private attr dictionary.

Up to now I've mostly focused on getting everything moved over to the new API and working again. Seeing as these are all private attributes (i.e. not user accessible) we can change the internal workings (e.g., by rearranging them in a dict) without breaking things.

src/ewatercycle/plugins/wflow/model.py
```diff
@@ -85,141 +58,79 @@ def _setup_default_config(self):
...
-        self, name: str, lat: Iterable[float], lon: Iterable[float]
-    ) -> Iterable[int]:
-        """Convert lat/lon values to index.
+        return self._work_dir
 
-        Args:
-            lat: Latitudinal value
-            lon: Longitudinal value
+    def get_parameters(self):
+        self._post_init()
```
Peter9192, line 154: this post_init is not defined, should it call `_parse_config` instead?

src/ewatercycle/plugins/hype/model.py
```diff
@@ -190,6 +165,20 @@ def get_value_as_xarray(self, name: str) -> xr.DataArray:
         """
         raise NotImplementedError("Hype coordinates cannot be mapped to grid")
 
+    def get_latlon_grid(self, name: str) -> tuple[Any, Any, Any]:
+        raise NotImplementedError("Hype coordinates cannot be mapped to grid")
+
+    def get_parameters(self) -> Iterable[Tuple[str, Any]]:
```
Peter9192, line 162: @BSchilperoort why did you change the parameters from a property called `parameters` to a method called `get_parameters`? Is this to do with what you said about pydantic?

Note that we want to use parameter akin to pymt: https://pymt.readthedocs.io/en/latest/usage.html#model-setup

BSchilperoort: I now see that `*Model.parameters` is defined as a variable property in `BaseModel`. I think it conflicted with defining the parameters as `@property` in the specific model implementations. 

I guess it's best to change how the BaseModel implements that.


Side note: I've found it a bit confusing that our `BaseModel` has the same class name as the pydantic `BaseModel` :confused: 
```py
class BaseModel(pydantic.BaseModel, abc.ABC):
    ...
```

src/ewatercycle/plugins/wflow/model.py
BSchilperoort: Oh yeah you're right. I guess the tests didn't catch any of that... I think renaming the `_parse_config` is best. 

Doing these operations in a `post_init` in all models would make them look more similar.

BSchilperoort:
> > Working with pydantic is a bit frustrating as it doesn't seem to fit our use completely.
> 
> We should definetely start using v2 asap. But if that's still frustrating, we could consider using "normal" dataclasses.

If we do at some point decide to not use pydantic, a good alternative is probably `typing`'s `Protocol`, and use `@runtime_checkable`. Although those checks are not as robust as Pydantic.

Peter9192:
Pff merge concluded.. 😅

sverhoeven (approved):
Better then before. Still lots to do.

Peter9192:
> Better then before. Still lots to do.

See #359

src/ewatercycle/plugins/hype/model.py
Peter9192: Renamed to eWaterCycleModel in https://github.com/eWaterCycle/ewatercycle/pull/361/commits/998b22684ae1a3e3f624ba97bd3fd282712d6ee9

Issue #359, opened 2023-07-27
Model baseclass follow-up todo's


Peter9192:
- [x] Update tests for abstractmodel to test the new API (once crystallized)
- [x] Update documentation
- [x] Remove `new_model_api.py` from repo.
- [x] Nice to have: Extend containerimage class to check version
- [ ] Nice to have: Extend containerimage class to automatically pull docker image and convert to singularity

and

- [x] iron out post_init / _parse_config (https://github.com/eWaterCycle/ewatercycle/pull/354#discussion_r1275998129)
- [x] reset `get_parameters` back to `parameters` property
- [ ] Version check: we could probaly reimplement an alternative version check validator on the eWaterCycle model

sverhoeven:
See https://pymt.readthedocs.io/en/latest/usage.html#model-setup for why we want bmi.parameters

Peter9192:
Remaining documentation issues will be addressed after the v2 release.

PR #361, opened 2023-07-27
Pydantic v2

### LINKED ISSUE #360: pydantic v2
https://docs.pydantic.dev/2.0/migration/

Peter9192:
refs #359 
closes #360 

Also adds version property to containerimage class

Todo:

- Clean up/refactor models to more standard layout (all config info stored in `_config` variable):
  - [x] marrmot
  - [x] pcrglobwb
  - [x] wflow
  - [x] hype
  - [x] lisflood


Peter9192:
I think the base classes should be mostly okay now, next step is fixing the plugins.

BSchilperoort:
image
@Peter9192, @sverhoeven should this be fixed as well? Is this a leftover issue due to the move to bmi 2?

Peter9192:
> Is this a leftover issue due to the move to bmi 2?

Hmm in some places we're using `OptionalDestBmi` so that `x` and `y` can remain optional here 🤔

Peter9192:
For both the local model https://github.com/eWaterCycle/ewatercycle/blob/572cd6f4b4ae76b5b3cb993a6e2003343146b22b/src/ewatercycle/base/model.py#L410-L410

and the containerized model https://github.com/eWaterCycle/ewatercycle/blob/572cd6f4b4ae76b5b3cb993a6e2003343146b22b/src/ewatercycle/container.py#L193

we're using `OptionalDestBmi`, but perhaps some of the type hints don't reflect that (yet)?

BSchilperoort:
> perhaps some of the type hints don't reflect that (yet)?

Or pylance detects this incorrectly. I don't see any problems when I run `mypy src/`

Peter9192:
Seeing the recent changes I wonder if we can move more config-related stuff to the base class. What do you think @BSchilperoort?

Peter9192:
Also I'm wondering whether we are misusing "protected" methods here, as we're overwriting them in most of our descendent classes:

https://github.com/eWaterCycle/ewatercycle/blob/572cd6f4b4ae76b5b3cb993a6e2003343146b22b/src/ewatercycle/base/model.py#L114-L116

We could convert them to public methods, or perhaps public (utility) functions (outside the class) that operate on instances of an `eWaterCycleModel`. Any thoughts @BSchilperoort @sverhoeven?

BSchilperoort:
I think the code that's left now is very model specific, as models have different config styles (xml, matlab .mat file formatted dict, etc.) and varying parameters. I feel like trying to generalize this would require something quite complex.

Of course we can make a default implementation that uses our preferred standard, which we can then use in the LeakyBucket implementation & documentation.

Peter9192:
Only remaining issue on sonarcloud has to do with abstractmodel tests (totally skipped for now). If we address that, we can also tick another box in #359. But perhaps do it in another PR.

src/ewatercycle/base/forcing.py
```diff
@@ -14,7 +14,7 @@
 FORCING_YAML = "ewatercycle_forcing.yaml"
 
 
-def _to_absolute_path(v):
+def _to_absolute_path(v: Union[str, Path]):
     return to_absolute_path(v)
```
BSchilperoort:
```suggestion
def _to_absolute_path(v: Union[str, Path]):
    """Wraps to_absolute_path to a single-arg function, to use as Pydantic validator.""" 
    return to_absolute_path(v)
```

BSchilperoort (approved):
> Only remaining issue on sonarcloud has to do with abstractmodel tests (totally skipped for now). If we address that, we can also tick another box in https://github.com/eWaterCycle/ewatercycle/issues/359. But perhaps do it in another PR.

Yes it's probably nice to round this PR off and work on that in a separate one.

I just noticed one thing that confused me a bit, but I think I figured it out (see suggestion).

src/ewatercycle/base/forcing.py
```diff
@@ -13,6 +14,10 @@
 FORCING_YAML = "ewatercycle_forcing.yaml"
 
 
+def _to_absolute_path(v: Union[str, Path]):
```
BSchilperoort, line 17:
```suggestion
def _to_absolute_path(v: Union[str, Path]):
    """Wraps to_absolute_path to a single-arg function, to use as Pydantic validator.""" 
```

PR #362, opened 2023-07-31
Use esmvaltool run recipe workaround for all plugins

sverhoeven:
Workaround was only implemented for wflow at https://github.com/eWaterCycle/ewatercycle/commit/0e972833e1b3df4ce5b8442cbdb2d3f91aedcc1c

TODO
* [x] `run_esmvaltool_recipe()`
* [x] wflow
  * [x] class
  * [x] notebook
* [x] pcrglobwb
  * [x] class
  * [x] notebook
* [x] marrmot
  * [x] class
  * [x] notebook
* [x] lisflood
  * [x] class
  * [x] notebook
* [x] hype
  * [x] class
  * [x] notebook


src/ewatercycle/base/forcing.py
```diff
@@ -150,6 +152,55 @@ def session_dir(self):
...
+        >>> recipe = get_recipe('hydrology/recipe_wflow.yml')
+        >>> recipe.data['scripts']['script']['dem_file'] = 'my_dem.nc'
+        >>> output_dir = Path('./output_dir')
+        >>> output = run_esmvaltool_recipe(recipe, output_dir)
+    """
+    # ESMVALCore 2.8.1 always runs original recipe,
+    # write updated recipe to disk and use
+    recipe.path = _write_recipe(recipe)
+    # TODO write recipe in output_dir?
+    # TODO fix in esmvalcore and wait for new version?
```
Peter9192, line 188: Indeed I think it's good to report the issue at esmvalcore and perhaps suggest a fix as well.

src/ewatercycle/plugins/marrmot/forcing.py
```diff
@@ -112,3 +121,62 @@ def generate(  # type: ignore
 
 # TODO could be nice to have plot function
 # that loads the mat file into a xarray dataset and plots it
+
+
+# TODO make method of MarrmotForcing class? What to call it?
```
Peter9192: Perhaps something like `to_xarray`? Something with "load" would be natural but it conflicts with the existing load method on the forcing class.

Peter9192 (approved):
Good improvements

Peter9192:
#361 has been merged which causes a few conflicts

PR #363, opened 2023-07-31
Make generic distributed esmvaltool recipe

sverhoeven:
The DefaultForcing now generates a generic distributed forcing.

TODO
* [ ] make sure template recipes is included in wheel


sverhoeven:
Switched to building recipes from scratch in #365, which is more flexible than editing a yaml file.

PR #377, opened 2023-09-14
List of BMI methods that this package or its dependency (grpc4bmi) ca…

sverhoeven:
…lls.


Peter9192 (approved):
Good to have this added somewhere

PR #382, opened 2023-10-02
The forcing directory should be required

sverhoeven:


BSchilperoort (approved):
lgtm.

I'll remove the assertion in the plugin template :+1:

PR #393, opened 2024-02-22
Allow for easier definition of new forcings

### LINKED ISSUE #232: Add `model.setup(cfg_dir=...)` example to user guide
Manually specifiying the model work dir is a useful feature, but it is hard to find since it's not explained in the user documentation. Can this be added as an example?

### LINKED ISSUE #387: Add Bart to CITATION.cff
@BSchilperoort contributed to repo but is not listed in https://github.com/eWaterCycle/ewatercycle/blob/main/CITATION.cff .

BSchilperoort:
Changes:

### Added
- Added new forcing classes ([#393](https://github.com/eWaterCycle/ewatercycle/pull/393)):
  - LumpedUserForcing, DistributedUserForcing - have the variable names as an argument, as well as an optional post-processor that can derive addition variables from the downloaded data.
  - LumpedMakkinkForcing, DistributedMakkinkForcing - based on the UserForcing: selects the required variables and computes the Makkink potential evaporation.
  - LumpedPenmanMonteithForcing, DistributedPenmanMonteithForcing - based on the UserForcing: includes all meteo variables required to compute the Penman-Monteith potential evaporation. More (parameterset) info such as plant type and height are required to compute the actual E_pot.

### Changed
- Internal changes to DefaultForcing: variables are no longer attributes of the class, but are stored under a "filenames" (dict) attribute ([#393](https://github.com/eWaterCycle/ewatercycle/pull/393)).
- Updated the documentation to reflect the changes in the forcing generation ([#393](https://github.com/eWaterCycle/ewatercycle/pull/393)).

I also tried to silence most useless esmvalcore warnings.

I split up the user_guide notebook into 4 parts, because it was too big to easily modify and run. They are now organized under a separate section on the docs, preserving their original structure.

A demo notebook is available here:
[_demo_forcing_builder.ipynb.zip](https://github.com/eWaterCycle/ewatercycle/files/14575533/_demo_forcing_builder.ipynb.zip)

Also closes #232, #387


docs/user_guide/03_models_obs_analysis.ipynb
sverhoeven, line 1: hype is listed in cell output, but missing in markdown.

src/ewatercycle/models.py
```diff
@@ -11,23 +11,89 @@
...
+from ewatercycle.base.model import eWaterCycleModel
 
 _model_entrypoints = entry_points(group="ewatercycle.models")  # /NOSONAR
 
 # Expose as "from ewatercycle.models import Model" for backward compatibility
 for _model in _model_entrypoints:
     globals()[_model.name] = _model.load()
+
+
+class ModelSources(Mapping):
```
sverhoeven: This class look very much like ForcingSource except the return type and its super is from typing vs collections.abc .  Can we have the same super?  Can we make a Source super class with a generic?

src/ewatercycle/util.py
```diff
@@ -137,6 +138,68 @@ def fit_extents_to_grid(extents, step=0.1, offset=0.05, ndigits=2) -> Dict[str,
...
 
 
+def merge_esvmaltool_datasets(datasets: list[xr.Dataset]) -> xr.Dataset:
+    """Merge the separate output datasets from an ESMValTool recipe into one dataset.
+
+    ESMValTool has bad management of floating point precision in coordinates. Every
+    CMORized file can have different rounding errors in the values of its coordinates.
+    This will prevent easy merging with xarray's open_mfdataset, or combine_by_coords.
+    By rounding to the 7th decimal place, more than sufficient precision is preserved
+    ('waldo-on-a-page' precision)[1], while solving the floating point inprecision
```
sverhoeven, line 148: did not know it was called `waldo-on-a-page` precision.

src/ewatercycle/forcing.py
```diff
@@ -42,12 +52,28 @@ def __len__(self):
...
 
 _forcings: dict[str, Any] = {
     "GenericDistributedForcing": GenericDistributedForcing,
     "GenericLumpedForcing": GenericLumpedForcing,
+    "DistributedUserForcing": DistributedUserForcing,
+    "LumpedUserForcing": LumpedUserForcing,
+    "DistributedMakkinkForcing": DistributedMakkinkForcing,
+    "LumpedMakkinkForcing": LumpedMakkinkForcing,
+    "DistributedPenmanMonteithForcing": DistributedPenmanMonteithForcing,
+    "LumpedPenmanMonteithForcing": LumpedPenmanMonteithForcing,
```
sverhoeven: Would be nice if the docstring of sources is updated to use one of the new forcing classses or gutted with link to user guide..

docs/user_guide/02_forcing.ipynb
sverhoeven, line 1: The links from the classes to api reference are not working. 

http://localhost:8000/_build/html/user_guide/02_forcing.html
has autoapi/ewatercycle/base/forcing/index.html#ewatercycle.base.forcing.GenericDistributedForcing
that link does not work, it should be to 
http://localhost:8000/_build/html/autoapi/ewatercycle/base/forcing/index.html#ewatercycle.base.forcing.GenericDistributedForcing

Adding `../` helps.

For classes inside `_forcings` module the deeplink (after fix) work, but they are not listed on the api reference. Can we rename the module to `forcings`

CITATION.cff
```diff
@@ -35,6 +35,11 @@ authors:
     family-names: Vreede
     given-names: Barbara
     orcid: "https://orcid.org/0000-0002-5023-4601"
+  -
+    affiliation: "Netherlands eScience Center"
+    family-names: Schilperoort
+    given-names: Bart
+    orcid: "https://orcid.org/0000-0003-4487-9822"
```
sverhoeven, line 42: Yeah

src/ewatercycle/_forcings/penman_monteith.py
```diff
@@ -0,0 +1,60 @@
...
+            postprocessor=None,
+            **model_specific_options,
+        )
+
+
+class LumpedPenmanMonteithForcing(PenmanMonteith, LumpedUserForcing):  # type: ignore[misc]
+    ...
+
+
+class DistributedPenmanMonteithForcing(PenmanMonteith, DistributedUserForcing):  # type: ignore[misc]
```
sverhoeven: I tried to run 

```
pmforcing = ewatercycle.forcing.sources.LumpedPenmanMonteithForcing.generate(
     dataset='ERA5',
    start_time="2000-01-01T00:00:00Z",
    end_time="2001-01-01T00:00:00Z",
    shape=shape.absolute(),
)   
print(pmforcing)
```

But got error:
ERROR:esmvalcore._recipe.recipe:Missing data for preprocessor diagnostic/rsus:
- Missing data for Dataset: rsus, day, OBS6, ERA5, 1
ERROR:esmvalcore._recipe.recipe:Missing data for preprocessor diagnostic/rlus:
- Missing data for Dataset: rlus, day, OBS6, ERA5, 1

This is with dcache mounted, to get the generate to work, those vars should be added to dcache.

If I try with cmip_dataset from user guide I get `OSError: [Errno 113] No route to host` error while LumpedMakkinkForcing worked this morning.

sverhoeven (changes requested):
The new forcing classes look great.

I was able to generate forcing using GenericDistributedForcing, LumpedMakkinkForcing,  LumpedUserForcing.

Initially LumpedMakkinkForcing.generate example gave`RecipeError: Could not create all tasks` with empty esmvaltool log files. When I disable the log swallowing in run_recipe(), the log files are filled again with esgf creds error. The log swallowing is a bit to much, now can dial it back.

In docs/adding_models.rst we should no longer just mention the Default and Generic forcings. It would be nice to link back to the forcing user guide.

Can you add tests for some of the new algorithms like sources, makkink and merge_esmvaltool_datasets?

The CI is red. Can you make it green?

docs/user_guide/03_models_obs_analysis.ipynb
BSchilperoort: Oh yeah. It was already missing from the original list and didn't compare the two. Well spotted.

src/ewatercycle/models.py
BSchilperoort: They are very similar yes. In some way it would be nicer to be able to list models, forcings, and parameter sets all in the same way. 

Before this change it was:
- `ewatercycle.forcing.sources`
- `dir(ewatercycle.models)` _which also was cluttered with non-useful info_
- `ewatercycle.parameter_sets.available_parameter_sets()`

It would also be nice to still unify how to access these... three different ways for what is essentially the same is not very user friendly.

BSchilperoort:
> Initially LumpedMakkinkForcing.generate example gave`RecipeError: Could not create all tasks` with empty esmvaltool log files. When I disable the log swallowing in run_recipe(), the log files are filled again with esgf creds error. The log swallowing is a bit to much, now can dial it back.

I'll change the warning level to "error" so actual errors are still shown, but useless warning are still hidden.
I find it a bit frustrating that esmvaltool just dumps everything to the terminal and if you quiet the logging it also doesn't write it to file :/

src/ewatercycle/_forcings/penman_monteith.py
BSchilperoort: I had another good look, and turns out that there's not too much overlap between the CMIP and ERA5 variables :roll_eyes: 

With some postprocessors I now have it working for both ERA5 and CMIP6.

> If I try with cmip_dataset from user guide I get OSError: [Errno 113] No route to host error while LumpedMakkinkForcing worked this morning.

I think ESGF had some issues. Should be working now (it runs fine for me now).

docs/user_guide/02_forcing.ipynb
BSchilperoort: I thought I'd tested these and they worked...

src/ewatercycle/models.py
BSchilperoort: > Can we have the same super? Can we make a Source super class with a generic?

Done.

> It would also be nice to still unify models, forcings, and parameter

I've now unified models and forcings, however parameter_sets it still tbd. Probably better in a separate PR.

sverhoeven: Would be nice if you did not need a subclass with __getitem__  method.
I tried `Mapping[str, DefaultForcing]` and [UserDict](https://docs.python.org/3/library/collections.html#userdict-objects) instead of Mapping, but got stuck on value being a class or an entrypoint.

On my machine it does not make much difference if load() is called or not:
```ipython
In [1]: from importlib.metadata import entry_points
In [4]: %timeit [f.load() for f in entry_points(group="ewatercycle.forcings")]
7.33 ms ± 158 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit entry_points(group="ewatercycle.forcings")
7.3 ms ± 81.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
So we could simplify by calling the load() on import.

docs/user_guide/02_forcing.ipynb
sverhoeven: Links in User-defined forcing chapter (https://ewatercycle--393.org.readthedocs.build/en/393/user_guide/02_forcing.html#User-defined-forcing) still give 404. Could you fix those as well?

sverhoeven (approved):
Looks good.

Only found 2 links in docs that are 404.

Peter9192:
Great progress!

src/ewatercycle/models.py
BSchilperoort: The subclassing here is to get the typing correct, not to delay the loading.

for the model sources:
```py
    def __getitem__(self, key) -> Type[eWaterCycleModel]:
```

while the forcing sources are:
```py
    def __getitem__(self, key) -> Type[DefaultForcing]:
```

It also allowed me to write a different docstring with appropriate examples for the model vs. forcing sources.

docs/user_guide/02_forcing.ipynb
BSchilperoort: fixed!

BSchilperoort:
I tested all models:

- [ ] leakybucket: needs update of plugin code (this PR breaks it)
- [ ] wflow: needs update in tests only (unexpected `filenames={}` in forcing object). model runs with old forcing.
- [ ] lisflood: needs update in tests only (unexpected `filenames={}` in forcing object). model runs with old forcing.
- [ ] hype: needs update in tests only (unexpected `filenames={}` in forcing object). Can't run model due to unrelated (known) issue https://github.com/eWaterCycle/ewatercycle-hype/issues/3
- [ ] pcrglobwb: needs update in tests only (unexpected `filenames={}` in forcing object). model runs with old forcing.
- [ ] marrmot: needs update in tests (unexpected `filenames={}` in forcing object). models run fine.

Merge + release should be OK. I'll open issues on the respective repos to track the tests breaking.

sverhoeven:
> I tested all models:
> 
> * [ ]  leakybucket: needs update of plugin code (this PR breaks it)
> * [ ]  wflow: needs update in tests only (unexpected `filenames={}` in forcing object). model runs with old forcing.
> * [ ]  lisflood: needs update in tests only (unexpected `filenames={}` in forcing object). model runs with old forcing.
> * [ ]  hype: needs update in tests only (unexpected `filenames={}` in forcing object). Can't run model due to unrelated (known) issue [GRPC server is down when running the `model.ipynb` ewatercycle-hype#3](https://github.com/eWaterCycle/ewatercycle-hype/issues/3)
> * [ ]  pcrglobwb: needs update in tests only (unexpected `filenames={}` in forcing object). model runs with old forcing.
> * [ ]  marrmot: needs update in tests (unexpected `filenames={}` in forcing object). models run fine.
> 
> Merge + release should be OK. I'll open issues on the respective repos to track the tests breaking.

Ah we are checking the dump of a forcing object in the tests, makes sense that those tests fail.

PR #396, opened 2024-03-14
Update system_setup.rst

sverhoeven:
PenmanMonteith forcing needs "rsus", "rlus" which are missing on dcache. 
Tried to follow instructions to download those, but they had to be fixed.

### DIFF
--- a/docs/system_setup.rst
+++ b/docs/system_setup.rst
@@ -83,7 +83,7 @@ To download ERA5 data files you can use the
 
     pip install era5cli
 
-Follow `instructions <https://era5cli.readthedocs.io/en/stable/instructions.html>`_ to get access to data.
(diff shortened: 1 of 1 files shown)


BSchilperoort:
Because of the lack of rsus/rlus in many CMIP datasets we decided to not make a PenmanMonteith forcing class. Instead we now have the Makkink forcing for which we only need the variables "pr", "tas", "rsds" which are already in the instructions.

docs/system_setup.rst
```diff
@@ -83,7 +83,7 @@ To download ERA5 data files you can use the
 
     pip install era5cli
 
-Follow `instructions <https://era5cli.readthedocs.io/en/stable/instructions.html>`_ to get access to data.
+Follow `instructions <https://era5cli.readthedocs.io/en/stable/getting_started/>`_ to get access to data.
```
BSchilperoort, line 86: This link fix is already in `main` now.

PR #403, opened 2024-03-29
Add missing model methods, documentation

### LINKED ISSUE #379: eWaterCycle BMI expects different get_time_units() vs. BMI 2.0
According to the BMI 2.0 specification, the function/method `get_time_units` should return the units of time like "s", "min", "h" or "d".

eWaterCycle expects the time units output to be parse-able to a datetime, like  `"seconds since 1970-01-01 00:00:00.0 +0000"`

### LINKED ISSUE #401: Clarify to users that an eWaterCycle model is not the same as BMI
Users don't talk directly to the BMI, but via the ewatercycle wrapper.
Some methods look the same, and it can give the user the impression they're using the BMI directly.

Some of the most commonly used methods have the same name:

- initialize
- update
- get_value
- set_value
- finalize

Others are very similar:

- get_value_at_coords (instead of get_value_at_indices)
- set_value_at_coords (instead of set_value_at_indices)
- start_time (instead of get_start_time)
- etc...

It should be very clear to users that `model.method` uses eWaterCycle-specific syntax, while `model.bmi.method` gives them access to underlying BMI methods (such as `get_var_units`)

@RolfHut

### LINKED ISSUE #402: Implement `model.var_units`
BMI has the method `.get_var_units`. In PyMT they use `.var_units(var_name)` in their wrapper.

eWaterCycle users have to do `model.bmi.get_var_units` which is not ideal.

BSchilperoort:
Closes #401 #402 

I've implemented the missing `var_units` and `input_var_names` model methods from PyMT.
The documentation is updated, to try to be a bit more clear about what methods are available (by linking to the API docs), and it's now explained how to access the _underlying_ bmi.

I also added a statement on the expected time_units to the model developer docs. Closes #379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Formatted as described on [https://keepachangelog.com](https://keepachangelog.co
 ### Added
 - Added the `input_var_names` property to the eWaterCycle model class, to accompany the existing `output_var_names` property ([#403](https://github.com/eWaterCycle/ewatercycle/pull/403)).
 - Added the `var_units` method to the eWaterCycle model class, to mirror PyMT ([#403](https://github.com/eWaterCycle/ewatercycle/pull/403)).
+- Added specification of the expected time units from the model's BMI to the documentation ([#403](https://github.com/eWaterCycle/ewatercycle/pull/403)).
 
 ### Changed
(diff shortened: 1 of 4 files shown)


RolfHut (approved):
looks good to me

BSchilperoort:
Thanks for the review!

PR #405, opened 2024-04-02
Add tests for the new ewatercycle model properties

BSchilperoort:
#403 was missing some unit tests. This PR adds them.


sverhoeven (approved):
Looks good. Testing on an empty tuple is not the most informative, but works.

Can you have a look at the pre-commit error? 
Having this branch and also the main branch have failing ci is not good.

BSchilperoort:
> Can you have a look at the pre-commit error? Having this branch and also the main branch have failing ci is not good.

Yeah I can fix it. There's a trailing whitespace on a .rst file

> Looks good. Testing on an empty tuple is not the most informative, but works.

Not the best, no. But the toy model doesn't have any input variables...

Issue #398, opened 2024-03-20
Add caravan dataset


Daafip:
'Caravan is an open community dataset of meteorological forcing data, catchment attributes, and discharge data for catchments around the world'.  [With a notable publication](https://www.nature.com/articles/s41597-023-01975-w) . 

Would be nice to easily access this as a standardised dataset in eWaterCycle. 
Currently you need to download a [12gb zip file](https://zenodo.org/records/7944025) and split the files yourself. 
Is there a way to nicely integrate this? 

'The Creative Commons Attribution license allows re-distribution and re-use of a licensed work on the condition that the creator is appropriately credited.' 
As far as I understand we are allowed to redistribute it?

BSchilperoort:
Nice that the dataset is available on Zenodo now. When I chatted with Frederik (the PI) last EGU, you basically had to go though Google Earth Engine to get the data.

It is a bit frustrating that everything is in a single zip file though. It would have been nice if the parts were more split up.

For integration with eWaterCycle I see one main hurdle:

> dataset of meteorological forcing data, catchment attributes, and discharge data

This combines eWaterCycle's separate forcing, parameter sets and observations in one block. So if you'd want to integrate this it would have to be split up into a `CaravanForcing`, a caravan `ParameterSet` definition, and a `caravan.py` module in observations.

Of course this is still completely viable.

We do not have a ZenodoDownloader implemented, but there is a placeholder [here](https://github.com/eWaterCycle/ewatercycle/blob/d02450bb079d213f1c7f2ac0100268b0ad8cab1b/src/ewatercycle/base/parameter_set.py#L95).

> As far as I understand we are allowed to redistribute it?

Yes. But we don't need to if we would implement a downloader. The users themselves would be downloading it.

Daafip:
The naming scheme is also different (to be expected), what worked for me to locally use camels files. 

-  `date` is used as dimension rather than `time`. 

```python
ds = ds.rename_dims({'date': 'time'})
ds = ds.rename({'date': 'time'})
```

- `pr`,`pev` etc. all have different names. To get HBV to run:

```python
RENAME_CAMELS = {'total_precipitation_sum':'pr',
                'potential_evaporation_sum':'pev',
                'streamflow':'Q'} 
ds = ds.rename(RENAME_CAMELS)
```

BSchilperoort:
[...]

PR #407, opened 2024-04-09
Adding caravan forcing

Daafip:
Hi All,

As this is my first contribution would like some feedback if possible. 

Was unsure on the naming convention, `generate` seems to be more linked to the esmvaltool, thus went for `retrieve`.

Runs for me when I run:

```py
from ewatercycle.forcing import sources
path = Path.cwd()
forcing_path = path / "Forcing"
experiment_start_date = "1997-08-01T00:00:00Z"
experiment_end_date = "2005-09-01T00:00:00Z"
HRU_id = 1022500

camels_forcing = sources['LumpedCaravanForcing'].retrieve(start_time = experiment_start_date,
                                                          end_time = experiment_end_date,
                                                          directory = forcing_path / "Camels",
                                                          basin_id = f"camels_0{HRU_id}",
                                                          variables = ('streamflow','potential_evaporation_sum'),
                                                          )
```
Discussions on the topic: #398 

Still todo: 
- [x] remove geopandas
- [x] remove wget
- [x] add documentation
- [x] add meaningfull example
- [x] Rename variables to correct ESMValTool 
- [x] Add test (using mocking)
- [x] fix failing test: `test_util.test_merge_esmvaltool_datasets`


Daafip:
Not too sure how to complete renaming from ERA-5 to ESMValTool compatible variable names. 
Do you have any insight on how to do that @BSchilperoort?

Edit: for now just supplied a rename dictionary as in discussion

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+from zipp import zipfile
+
+from ewatercycle.base.forcing import DefaultForcing, LumpedUserForcing
+from ewatercycle.util import get_time
+
+COMMON_URL = "ca13056c-c347-4a27-b320-930c2a4dd207"
+OPENDAP_URL = f"https://opendap.4tu.nl/thredds/dodsC/data2/djht/{COMMON_URL}/1/"
+SHAPEFILE_URL = f"https://data.4tu.nl/file/{COMMON_URL}/bbe94526-cf1a-4b96-8155-244f20094719"
+
+PROPERTY_VARS = ['timezone',
```
BSchilperoort: The code formatter doesn't like this.

You can run it with "pre-commit":
- `pip install pre-commit`
- `pre-commit run -a`

Usually it only looks at staged files, but with the `-a` flag you can run it on all files. This should also fix the CI.

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+        camels_forcing = sources['LumpedCaravanForcing'].retrieve(start_time = experiment_start_date,
+                                                                  end_time = experiment_end_date,
+                                                                  directory = forcing_path / "Camels",
+                                                                  basin_id = f"camels_0{HRU_id}"
+                                                                )
+        which gives somthing like:
+
+        .. code-block:: python
+
+            LumpedCaravanForcing(
```
BSchilperoort: It looks a lot like the other forcings now, which is nice. That should make it easier to swap this forcing with, e.g. Makkink, or the other way around.

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+                start_time='1997-08-01T00:00:00Z',
+                end_time='2005-09-01T00:00:00Z',
+                directory=PosixPath('/home/davidhaasnoot/eWaterCycle-WSL-WIP/Forcing/Camels'),
+                shape=PosixPath('/home/davidhaasnoot/eWaterCycle-WSL-WIP/Forcing/Camels/shapefiles/camels_01022500.shp'),
```
BSchilperoort: I recently changed the code so the shapefile is copied along and stored next to the forcing files. 

That is part of the `.save` method. I think it works slightly differently here 🤔 . Or does it work the exact same?

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+                    'evspsblpot': 'camels_01022500_1997-08-01T00:00:00Z_2005-09-01T00:00:00Z_evspsblpot.nc',
+                    'pr': 'camels_01022500_1997-08-01T00:00:00Z_2005-09-01T00:00:00Z_pr.nc',
+                    'tas': 'camels_01022500_1997-08-01T00:00:00Z_2005-09-01T00:00:00Z_tas.nc',
+                    'Q': 'camels_01022500_1997-08-01T00:00:00Z_2005-09-01T00:00:00Z_Q.nc'
+                }
+            )
+    """
+
+    @classmethod
+    def retrieve(cls: Type["Caravan"],
```
BSchilperoort: I am in doubt about `retrieve` vs. `generate`...

`generate` would be better for consistency.
`retrieve` is more descriptive of what this method does.

At the least, you perhaps should make `.generate` raise a better error, at the moment you don't overwrite the default forcing's method even though that does contain an implementation.

BSchilperoort: @Peter9192 what is your opinion on this?

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+        dataset = basin_id.split('_')[0]
+        ds = xr.open_dataset(
+            f"{OPENDAP_URL}{dataset}.nc")
+        ds_basin = ds.sel(basin_id=basin_id.encode())
+        ds_basin_time = crop_ds(ds_basin, start_time, end_time)
+
+        if shape is None:
+            shape = get_shapefiles(Path(directory), basin_id)
+
+        if 'lat' not in variables and 'lon' not in variables:
```
BSchilperoort: Did some of the basins miss lat and lon values? I expected all of them to have a proper value... Otherwise we should fix the input data.

It'd be better if it was correct there already instead of needing to fix it here.

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+            shape_obj = shapereader.Reader(shape)
+            centre = next(shape_obj.geometries()).centroid
+            ds_basin_time.coords.update({'lat': centre.y,
+                                         'lon': centre.x
+                                         })
+        if variables == ():
+            variables = (ds_basin_time.data_vars.keys())
+
+        # only return the properties which are also in property vars
+        properties = tuple(set(variables).intersection(set(PROPERTY_VARS)))
```
BSchilperoort: I love python's sets 😄

BSchilperoort: I don't think you need it to be a tuple. You can iterate over a set as well.
Also, I believe `.intersection()` turns the argument into a set already.
```suggestion
        properties = set(variables).intersection(PROPERTY_VARS)
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+        variables = tuple(
+            set(variables).difference(set(PROPERTY_VARS)).intersection(
+              set(RENAME_ERA5.keys())
+            )) # only take the vars also in Rename dict
```
BSchilperoort: I think this should do the same, but is more readable:
```suggestion
        non_property_vars = set(variables) - properties
        variable_names= non_propery_vars.intersection(RENAME_ERA5.keys())
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+
+        for prop in properties:
+            ds_basin_time.coords.update({prop: ds_basin_time[prop].to_numpy()})
+
+        ds_basin_time = ds_basin_time.rename(RENAME_ERA5)
+        variables = ([RENAME_ERA5[var] for var in variables])
```
```suggestion
        variables = ([RENAME_ERA5[var] for var in variable_names])
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+            shape=Path(shape),
+            filenames={
+                var: f'{basin_id}_{start_time}_{end_time}_{var}.nc' for var in variables
+            }
+        )
+        forcing.save()
+        return forcing
+
+
+class LumpedCaravanForcing(Caravan, LumpedUserForcing):  # type: ignore[misc]
```
BSchilperoort: You can move this to the top. There won't be a distributed caravan forcing, so you can just inherit in the class containing the implementations.

Also, as there won't be a DistributedCaravanForcing, it might be better to just name it `CaravanForcing`.

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+
+def get_shapefiles(directory: Path, basin_id: str):
+    """Retrieves shapefiles from data 4TU."""
+    zip_path = directory / 'shapefiles.zip'
+    output_path = directory / 'shapefiles'
+
+    if not zip_path.is_file():
+        timeout = 30
+        try:
+            result = requests.get(SHAPEFILE_URL, timeout=timeout)
```
BSchilperoort: You should wrap this in a context manager so `.close` is called on the response.

```py
with requests.get(...) as result:
    ...
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,240 @@
...
+
+    shape_obj = shapereader.Reader(combined_shapefile_path)
+    list_records = []
+    for record in shape_obj.records():
+        list_records.append(record.attributes['gauge_id'])
+
+    df = pd.DataFrame(data=list_records, index=range(len(list_records)), columns=['basin_id'])
+    basin_index = df[df['basin_id'] == basin_id].index.array[0]
+
+    with fiona.open(
```
BSchilperoort: Perhaps you can refactor this to a separate function to make the main function more readable.

BSchilperoort (changes requested):
Hi David, nice work! I left some comments for possible improvements to the code. Feel free to ask a review again when it's ready!

BSchilperoort:
Also, it would be nice if you could add tests for this forcing. (without downloading anything). If you're unsure on how to do that we can discuss it at some point

src/ewatercycle/_forcings/caravan.py
BSchilperoort: See https://github.com/WAM2layers/WAM2layers/pull/368#discussion_r1559049999 for some more in-depth info

src/ewatercycle/_forcings/caravan.py
Daafip: Yup, might also be nice to also have the seperate Camels data sets as these do vary slightly. (But thats more work for another time)

src/ewatercycle/_forcings/caravan.py
Daafip: Good point, will change that

src/ewatercycle/_forcings/caravan.py
Daafip: They all have lat/lon, was my first way of fixing the issue. Then i just moved the variables. Will remove the redundant code.

src/ewatercycle/_forcings/caravan.py
Daafip: Thanks, that cleans it up nicely

RolfHut:
regarding conventions: from a didactic point of view in tutorial notebooks I suggest to always use:
```python
import ewatercycle.forcing

camels_forcing = ewatercycle.forcing.sources['LumpedCaravanForcing'].retrieve(start_time = experiment_start_date,
                                                          end_time = experiment_end_date,
                                                          directory = forcing_path / "Camels",
                                                          basin_id = f"camels_0{HRU_id}",
                                                          variables = ('streamflow','potential_evaporation_sum'),
                                                          )
```
Novice users quickly loose sight of which function belongs to which library when using ```from ewatercycle.forcing import sources```. This might load a bit more than you would ideally want, but communicates more clearly. In code 'under the hood' I'm more than fine with the shorter version.

src/ewatercycle/_forcings/caravan.py
Daafip: Made `.generate` now raises an `NotImplementedError` directing you to `retrieve`.
Retrieve makes more sense to me as it also makes it clear to the user it's an already prepared dataset. 
But interested to see what Peter says.

src/ewatercycle/_forcings/caravan.py
Daafip: Still getting issues with CI:
```
nbqa-isort...............................................................Failed
..
ModuleNotFoundError: No module named 'pkg_resources'
```
 and 
```
nbqa-flake8..............................................................Failed
...
AttributeError: 'EntryPoints' object has no attribute 'get'
```

Daafip:
Made the suggested changes Rolf, Example Notebook can be found here: 
https://gist.github.com/Daafip/ac1b030eb5563a76f4d02175f2716fd7

Daafip:
Still facing three issues, maybe you could help with that Bart?: 
- [x] test_utipl.test_merge_esmvaltool_datasets fails: . If i run the test locally in my IDE I don't have this problem but on the git tests I Do...? 
```
FAILED tests/src/test_util.py::test_merge_esmvaltool_datasets - AssertionError: assert 'height' in {'cell_methods': 'day_of_year: year: mean', 'long_name': 'Near-Surface Air Temperature', 'standard_name': 'air_temperature', 'units': 'K'}
```
- [x] `MyPy` throws an error I dont understand in the SonarCloud analysis:
```
mypy.....................................................................Failed
-hook id: mypy
-exit code: 1

src/ewatercycle/_forcings/caravan.py:10: error: Library stubs not installed for "requests"  [import]
src/ewatercycle/_forcings/caravan.py:10: note: Hint: "python3 -m pip install types-requests"
src/ewatercycle/_forcings/caravan.py:10: note: (or run "mypy --install-types" to install all missing stub packages)
src/ewatercycle/_forcings/caravan.py:10: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```
- [x] Not sure how to make tests not download

BSchilperoort:
> test_utipl.test_merge_esmvaltool_datasets fails: . If i run the test locally in my IDE I don't have this problem but on the git tests I Do...?

That's odd. The "height" attribute should be in there. Perhaps it's an environment/versioning issue, and the environment is newly created on Github Actions for the test. So you don't see the error as you don't have all the latest versions of packages.

> MyPy throws an error I dont understand in the SonarCloud analysis:

Probably also due to a version change. You can add the `types-request` package to the setup.cfg file below `types-PyYAML`. That should fix it.

> Not sure how to make tests not download

You would have to mock/patch some of the code. For examples see https://github.com/eWaterCycle/ewatercycle/blob/b4ea453727aef571dbc18459d1804c8f86c75ff0/tests/src/test_container.py#L177

You'd mock this line in your code `ds = xr.open_dataset(f"{OPENDAP_URL}{dataset}.nc")`, with `mock.patch("xarray.Dataset') as ...`.
You can then set the return value to return a dataset you load from a small testing file. And assert that the mocked function was called with the correct arguments.

Daafip:
> That's odd. The "height" attribute should be in there. Perhaps it's an environment/versioning issue, and the environment is newly created on Github Actions for the test. So you don't see the error as you don't have all the latest versions of packages.

Still dont really understand why this fails, I think it might be a versioning issue, played around a bit in a notebook (https://gist.github.com/Daafip/3a4f8650f499e9a2b68fc187ef7d73d7) and it only worked if I loaded the function seperately. This is in a new environment using the eWaterCycle/ewatercycle source code and the latest pip release.  Unsure how to fix this though..

> SonarCloud analysis

Fixed mypy, now have an issue with the SonarCloud token, but i can't acces the repo settings..

"Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork." - any way to fix this? 

> You would have to mock/patch some of the code. For examples see

 Ah okay, Will look at mokking tests.

src/ewatercycle/_forcings/caravan.py
Peter9192: From reading the conversation only I think I'm okay with the proposed solution :-)

Daafip:
Added a basic test now, but took me quite a bit of time. 
Would you have time to take a look at the issues im having with SonarCloud & the other test @BSchilperoort?

RolfHut:
apart from the technical comments, a more hydrological point: we do have to make sure in the documentation that people are aware that the Caravan dataset is not "just" the Camels grouped together, but rather it is new (ERA) data derived for the shape files in the Camels datasets. See https://egusphere.copernicus.org/preprints/2024/egusphere-2024-864/ for a ongoing discussion in the hydrological community about this.

BSchilperoort:
> Added a basic test now, but took me quite a bit of time. Would you have time to take a look at the issues im having with SonarCloud & the other test @BSchilperoort?

Fixed the isort and black issues. I now see that there's the "end-of-file-fixer" which (incorrectly!) is trying to fix those shapefile file. I did not find any way to configure it, so I removed it from the pre-commit setup. Black already adds a newline to the end of .py files anyway.

tests/src/base/test_forcing.py
```diff
@@ -270,3 +276,54 @@ def recipe_output_cls(cls, *args, **kwargs):
             .isnull()
             .any("time")
         )
+
+
+@pytest.fixture
```
BSchilperoort, line 282: I don't think this mock is not being used below. Have a look at the test_container.py file.

The fixture needs to be an input argument for the test function.

By adding `mocked_thing.assert_called_once_with(...` you can make sure that the mock was correctly used.

BSchilperoort:
Sonarcloud is still failing due to "Set the SONAR_TOKEN env variable.". However, in the repo's secrets this is correctly set and it used to work fine.

No idea how to fix this. There doesn't seem to be a new version out so it's completely unclear to me why the workflow stopped working. Three weeks ago the it still ran successfully. 

Any ideas @sverhoeven ?

Daafip:
> Sonarcloud is still failing due to "Set the SONAR_TOKEN env variable.". However, in the repo's secrets this is correctly set and it used to work fine.
> 
> No idea how to fix this. There doesn't seem to be a new version out so it's completely unclear to me why the workflow stopped working. Three weeks ago the it still ran successfully.
> 
> Any ideas @sverhoeven ?

I think it might be because I forked, can you create a new branch in the repo (i dont have the rights to do so) and see if that fixes it?

tests/src/base/test_forcing.py
Daafip: Ah, will look into this

BSchilperoort:
> > Sonarcloud is still failing due to "Set the SONAR_TOKEN env variable.". However, in the repo's secrets this is correctly set and it used to work fine.
> > No idea how to fix this. There doesn't seem to be a new version out so it's completely unclear to me why the workflow stopped working. Three weeks ago the it still ran successfully.
> > Any ideas @sverhoeven ?
> 
> I think it might be because I forked, can you create a new branch in the repo (i dont have the rights to do so) and see if that fixes it

I think you're completely correct with that! Good thinking :+1:

I won't bother making a new branch on this repo, pretty sure that's the cause of the problem. I have given you write access though, so for a next PR you can make a branch on the repo.

tests/src/base/forcing_files/test_caravan_file.nc
BSchilperoort, line 1: If these are real Caravan netCDF and shape files it might be best to add a readme with a license statement.

tests/src/base/test_forcing.py
Daafip: It indeed wasn't actually mocking it properly using  `with mock.patch("xarray.open_dataset")`
If i use `with mock.patch("xarray.open_dataset")...`: it calls it correctly and opens the dataset, but also mocks `open_dataset` in the `caravan_forcing.to_xarray()` which then break things...
Will keep trying to fix it

Daafip: ah fixed it, https://changhsinlee.com/pytest-mock/ this is a useful resource

Daafip:
> That's odd. The "height" attribute should be in there. Perhaps it's an environment/versioning issue, and the environment is newly created on Github Actions for the test. So you don't see the error as you don't have all the latest versions of packages.

Still don't understand why this test fails...

Otherwise I think its mostly there

sverhoeven:
Can you run pre-commit locally? `pip install pre-commit` + `pre-commit run --all-files`, you can ignore the flake8 warnings, but errors should be fixed.

.pre-commit-config.yaml
```diff
@@ -7,7 +7,6 @@ repos:
       - id: check-case-conflict
       - id: check-merge-conflict
       - id: debug-statements
-      - id: end-of-file-fixer
```
sverhoeven, line 10: Instead of removing hook can shape files be excluded 
see https://pre-commit.com/index.html#regular-expressions

Daafip: thanks, fixed that now. Still failing on the test but once we mere that PR, should be fixed

BSchilperoort:
Hi David, we found the test issue and fixed it in #410 . Once that's merged to main that should resolve the problem you encounter here.

On second thought I do think that it would be better to use the `generate` method to generate the forcing. You can add extra kwargs (i.e. basin_id, and just leave the `shape` one unused).

From the converage output in the CI I see this;
```
src/ewatercycle/_forcings/caravan.py                   85     28     32      3    60%
```

Generally we want 80% coverage on the code (sonarcloud analysis). I think if you add tests for the `extract_basin_shapefile` function that you might be there already.

After these things I think this PR is ready to merge! :+1:

Daafip:
> I think if you add tests for the extract_basin_shapefile

Done! That Bumps it up to 83%
`src/ewatercycle/_forcings/caravan.py                   84     11     32      3    83%`

> On second thought I do think that it would be better to use the generate method to generate the forcing. You can add extra kwargs (i.e. basin_id, and just leave the shape one unused).

Refctored this now, so all ready! 

Pytest fails on the other test, sonar cloud is stuck on the token but should both be fine.

Thanks for all the input & feedback @BSchilperoort @sverhoeven!

Daafip:
> apart from the technical comments, a more hydrological point: we do have to make sure in the documentation that people are aware that the Caravan dataset is not "just" the Camels grouped together, but rather it is new (ERA) data derived for the shape files in the Camels datasets. See https://egusphere.copernicus.org/preprints/2024/egusphere-2024-864/ for a ongoing discussion in the hydrological community about this.

To get back to you on this  @RolfHut, In the data set attributes under history it does mention the data source. This is now correctly copied over to the new datas set. If the user looks at the (meta)data, they should see its ERA5

src/ewatercycle/_forcings/caravan.py
```diff
@@ -123,15 +124,21 @@ def generate(
...
                 If none is specified, will be downloaded automatically.
-            **kwargs
+            dataset: Unused
+
+            **kwargs:
+                basin_id: str containing the wanted basin_id.
+
         """
+        assert type(dataset) == str
+        basin_id = kwargs["basin_id"]
```
BSchilperoort, line 144: Can you add a check that gives the user a clear input error if the basin ID is missing?

BSchilperoort: Once this is fixed and you've merged from main (failing test is fixed in there), I'll approve and merge the PR.

Can you also update the changelog?

Daafip: Ah forgot to remove line 139, the `basin_id` is now passed as a kwarg rather than using the dataset variable. 

Will merge from main and update changelog

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Formatted as described on [https://keepachangelog.com](https://keepachangelog.co
 - Added the `input_var_names` property to the eWaterCycle model class, to accompany the existing `output_var_names` property ([#403](https://github.com/eWaterCycle/ewatercycle/pull/403)).
 - Added the `var_units` method to the eWaterCycle model class, to mirror PyMT ([#403](https://github.com/eWaterCycle/ewatercycle/pull/403)).
 - Added a note on the time units that eWaterCycle expects models to provide to the documentation ([#403](https://github.com/eWaterCycle/ewatercycle/pull/403)).
+- Added LumpedCaravan forcing, allows retrival of single basin from 4tu.nl rather than downloading the whole dataset. ([#407](https://github.com/eWaterCycle/ewatercycle/pull/407))
```
BSchilperoort:
```suggestion
- Added LumpedCaravan forcing, which allows for the retrieval of single basins from the Caravan dataset, rather than downloading the whole dataset ([#407](https://github.com/eWaterCycle/ewatercycle/pull/407)).
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -136,7 +136,6 @@ def generate(  # type: ignore[override]
                 basin_id: str containing the wanted basin_id.
 
         """
-        assert type(dataset) == str
         basin_id = kwargs["basin_id"]
```
BSchilperoort, line 144:
```suggestion
        if "basin_id" not in kwargs:
            msg = "You have to specify a basin ID to be able to generate forcing from Caravan."
            raise InputError(msg)
        basin_id = kwargs["basin_id"]
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -0,0 +1,270 @@
...
+            directory: Directory in which forcing should be written.
+                If not given will create timestamped directory.
+            variables: Variables which are needed for model,
+                if not specified will default to all.
+            shape: (Optional) Path to a shape file.
+                If none is specified, will be downloaded automatically.
+            dataset: Unused
+
+            **kwargs:
+                basin_id: str containing the wanted basin_id.
```
BSchilperoort: Can you add where the user can lookup the basin ID?

BSchilperoort (approved):
Nice work! Sorry that it was blocked a bit by the failing test+sonarcloud problems, but now it's ready to merge (after that 1 last comment above, and pre-commit haha) :smile: 

Edit: InputError isn't a buildin exception. You can pick a different appropriate one. This is why mypy fails.

src/ewatercycle/_forcings/caravan.py
Daafip: See this is where geopandas would be nice :p, but will add a function which lets the user explore the dataset(ids).

BSchilperoort: I was hoping more for something like a webpage listing the basin_id's together with river/basin names, country, etc. I guess that doesn't exist?

Daafip: Nope, but what i do now is list the datasets in a seperate function. That way the user can get the dataset as netcdf and a list of basin ID's

Daafip: For advanced users this will do, for more novice users choosing a single catchment: likely they will have to download the combinedshapefile, load this into GIS and pick one

Daafip: > I was hoping more for something like a webpage listing the basin_id's together with river/basin names, country, etc. I guess that doesn't exist?

would be pretty easy to make an interactive folium/leaflet map out of this actually! Similar to [this](https://nbviewer.org/github/Daafip/KNMI-Folium-Map/blob/main/Hourly%20%26%20daily%20station%20data%20KNMI%202023-03-21.html)

Daafip: Lets continue this discussion in issues: #398

PR #410, opened 2024-04-29
Fix merge_esmvaltool_datasets function

BSchilperoort:
The merge_esmvaltool_datasets function underlying the generic forcing's `.to_xarray()` method was not very robust. This lead to issues in the CI and #407 .

This PR addresses those problems.


sverhoeven (approved):
Looks good. Has everything we talked about.

Did you test if the `ds.copy(deep=True)` explodes the memory usage?

Would be nice to see unit tests for the removal of _bnds var and/or the ValueError.

BSchilperoort:
> Did you test if the ds.copy(deep=True) explodes the memory usage?

It would increase memory usage if the datasets aren't loaded in lazily. I changed the code to ensure that xr.Datasets with dask arrays are passed to the merger function, this should avoid that.

> Would be nice to see unit tests for the removal of _bnds vars

The bounds vars do end up in the final dataset (but only one version, as they differ by single precision floating point errors between the different emsvaltool output datasets). I now test if they are present in the final data.

> and/or the ValueError.

This is now properly tested :+1:

BSchilperoort:
Also, dask is now specified as a dependency. This should be an issue as we already depended on it through esmvaltool, but now it's more explicit.

PR #412, opened 2024-05-02
change InputError to KeyError

Daafip:
Missed the Bart's last comment in #407 :||, this works better if we can actually see the sonar could test

### DIFF
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -303,8 +303,8 @@ def extract_basin_shapefile(
 
 def crop_ds(ds: xr.Dataset, start_time: str, end_time: str) -> xr.Dataset:
     """Crops dataset based on time."""
-    start = pd.Timestamp(get_time(start_time))
-    end = pd.Timestamp(get_time(end_time))
+    start = pd.Timestamp(get_time(start_time)).tz_convert(None)
+    end = pd.Timestamp(get_time(end_time)).tz_convert(None)
     return ds.isel(
         time=(ds["time"].to_numpy() >= start) & (ds["time"].to_numpy() <= end)
     )
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -107,6 +107,45 @@ class CaravanForcing(DefaultForcing):
     """
 
     @classmethod
+    def get_dataset(cls: Type["CaravanForcing"], dataset: str) -> xr.Dataset:
+        """Opens specified dataset from data.4tu.nl OPeNDAP server.
+
+        Args:
+            dataset (str): name of dataset, choose from:
+                'camels',
+                'camelsaus',
+                'camelsbr',
+                'camelscl',
+                'camelsgb',
+                'hysets',
+                'lamah'
+        """
+        return xr.open_dataset(f"{OPENDAP_URL}{dataset}.nc")
+
+    @classmethod
+    def get_basin_id(cls: Type["CaravanForcing"], dataset: str) -> list[str]:
+        """Gets a list of all the basin ids in provided dataset
+        Args:
+            dataset (str): name of dataset, choose from:
+                'camels',
+                'camelsaus',
+                'camelsbr',
+                'camelscl',
+                'camelsgb',
+                'hysets',
+                'lamah'
+
+        Note:
+            https://www.ewatercycle.org/caravan-map/ contains online a set of
+            interactive maps which allows exploration of the available catchments and
+            also contains the needed basin_ids.
+            Alternatively, a zip with shapefiles is available at
+            https://doi.org/10.4121/ca13056c-c347-4a27-b320-930c2a4dd207.v1 which also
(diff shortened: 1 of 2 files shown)


Daafip:
> I was hoping more for something like a webpage listing the basin_id's together with river/basin names, country, etc. I guess that doesn't exist?

Shall I also add a link to these maps with the basin_ids? https://github.com/Daafip/caravan-map

BSchilperoort:
I was also working on fixing some things yesterday, see https://github.com/eWaterCycle/ewatercycle/commit/876efe8f693e8afe58e28e28b61b5a56f89480ae
Perhaps you can merge those changes. I was making the `get_basin_id` function a classmethod, as that's how it was described in the docstring.

> Shall I also add a link to these maps with the basin_ids? https://github.com/Daafip/caravan-map

Could be nice, yes. But perhaps better if you can host it on github pages. Perhaps moving it to the ewatercycle org might also be better. Also see the issue I opened https://github.com/Daafip/caravan-map/issues/1

Daafip:
> But perhaps better if you can host it on github pages. Perhaps moving it to the ewatercycle org might also be better

https://www.ewatercycle.org/caravan-map/ Done! 

Do you have any suggestions for the index page?

BSchilperoort:
> Do you have any suggestions for the index page?

I think if you copy part of the README there that would be good. Some explanation why the map exists (to explore and find basin IDs). And put in some relevant links to e.g. the ewatercycle documentation.

Daafip:
> function a classmethod

How do I then mock  the function if its a class method?

BSchilperoort:
> > function a classmethod
> 
> How do I then mock the function if its a class method?

I believe just with [mock.patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch).
https://stackoverflow.com/a/53485182

Daafip:
> > > function a classmethod
> > 
> > 
> > How do I then mock the function if its a class method?
> 
> I believe just with [mock.patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch). https://stackoverflow.com/a/53485182

Ah, issue with my local unit testing and having the correct version

Daafip:
Tried to fix the test coverage but don't think a higher coverage is easily met as most of uncovered is downloading files..

src/ewatercycle/_forcings/caravan.py
```diff
@@ -300,8 +305,8 @@ def extract_basin_shapefile(
 
 def crop_ds(ds: xr.Dataset, start_time: str, end_time: str) -> xr.Dataset:
     """Crops dataset based on time."""
-    get_time(start_time), get_time(end_time)  # if utc, remove Z to parse to np.dt64
-    start, end = np.datetime64(start_time[:-1]), np.datetime64(end_time[:-1])
+    start = pd.Timestamp(start_time[:-1])
```
BSchilperoort: Perhaps it's more robust to use `get_time` from `ewatercycle.util` :thinking:

Also you don't need to remove the "Z" (Zulu = UTC+0). For pandas pd.Timestamp. However, then you get a timezone aware timestamp which can be more difficult.

tests/src/base/test_forcing.py
```diff
@@ -331,6 +334,45 @@ def test_retrieve_caravan_forcing(tmp_path: Path, mock_retrieve: mock.MagicMock)
...
+
+def test_retrieve_caravan_forcing_no_basin_id(
+    tmp_path: Path, mock_retrieve: mock.MagicMock
+):
+    test_files_dir = Path(__file__).parent / "forcing_files"
+    tmp_camels_dir = tmp_path / "camels"
+    copytree(test_files_dir, tmp_camels_dir)
+
+    msg = (
+        "You have to specify a basin ID to be able to generate forcing from" " Caravan."
```
BSchilperoort:
```suggestion
        "You have to specify a basin ID to be able to generate forcing from Caravan."
```
unnecessary string concatenation

BSchilperoort (approved):
Just two minor comments. Feel free to merge once these are resolved.

With the downloading writing sensible tests can be difficult. But I think this file is sufficiently covered now. The file has an 85% coverage, sonarcloud just doesn't like that some of the changed lines in this PR aren't covered (the measure is per PR, not per file...)

src/ewatercycle/_forcings/caravan.py
Daafip: Was having issue with timezone aware mismatch...
If you do `pd.Timestamp(starttime)..`, the object you create is timezone aware whilst the forcing data is not...
Ill go with the get_time

BSchilperoort: I thought get_time would you give a timezone unaware datetime but not it's going wrong anyway? :/

Sorry for the confusion. I guess the original solution was quite OK, but it's probably good to parse out the "Z" differently, e.g., with `.replace("Z","")`.

Daafip: I went with `pd.Timestamp(get_time(start_time)).tz_convert(None)`
Makes it clear timezone awareness is dropped

PR #416, opened 2024-05-23
fixing bug, now takes only history attr

Daafip:
Copied over reserved variables. Now just takes the wanted variable.

See https://github.com/Unidata/netcdf4-python/issues/1020


BSchilperoort:
> Copied over reserved variables

strange how it copies reserved names which it doesn't show :/

```
>>> ds.attrs
{'history': 'Wed Mar 27 16:11:00 2024: /usr/bin/ncap2 -s time=double(time) -O Caravan/camels.nc Caravan/camels.nc\nMerged together from separate files; All forcing and state variables are derived from ERA5-Land hourly by ECMWF. Streamflow data was taken from the CAMELS (US) dataset by Newman et al. (2014).', 'NCO': 'netCDF Operators version 5.0.6 (Homepage = http://nco.sf.net, Code = http://github.com/nco/nco)'}
```
Nothing weird here, but I guess there is some stuff hidden?

BSchilperoort (approved):
Thanks for fixing this. I assume you ran Rolf's notebook successfully after this change? If so feel free to merge after updating the changelog.

Daafip:
>I assume you ran Rolf's notebook successfully after this change? 

yup https://gist.github.com/Daafip/8fb62aea4abba00c938a0b117a25b9d3

Daafip:
> Nothing weird here, but I guess there is some stuff hidden?

`_NCProperties`  to be precise, which is a reserved attribute name, so if you pass that to a new dataset it doesn't like it

PR #417, opened 2024-05-29
Change units caravan forcing

RolfHut:
I have added a few lines to change the units of the caravan forcings to be in line with the NETCDF CF conventions and compatible with the output of the Makking and Generic Lumped forcings.

### DIFF
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -209,7 +209,7 @@ class CaravanForcing(DefaultForcing):
 
         for temp in ["tas", "tasmin", "tasmax"]:
             ds_basin_time[temp].attrs.update({"height": "2m"})
-            #convert units to Kelvin for compatiabillity with NetCDF-CF conventions
+            #convert units to Kelvin for compatibility with CMOR MIP table units
             if (ds_basin[temp].attrs["unit"]) == "°C":
                 ds_basin[temp].values = ds_basin[temp].values + 273.15
                 ds_basin[temp].attrs["unit"] = "°K"
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -210,11 +210,15 @@ class CaravanForcing(DefaultForcing):
         for temp in ["tas", "tasmin", "tasmax"]:
             ds_basin_time[temp].attrs.update({"height": "2m"})
             #convert units to Kelvin for compatiabillity with NetCDF-CF conventions
-            ds_basin_time[temp].pint.to(a=temp, u = "K")
+            if (ds_basin[temp].attrs["unit"]) == "°C":
+                ds_basin[temp].values = ds_basin[temp].values + 273.15
+                ds_basin[temp].attrs["unit"] = "°K"
         
         for var in ["evspsblpot", "pr"]:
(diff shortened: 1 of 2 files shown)


src/ewatercycle/_forcings/caravan.py
```diff
@@ -209,6 +209,16 @@ def generate(  # type: ignore[override]
 
         for temp in ["tas", "tasmin", "tasmax"]:
             ds_basin_time[temp].attrs.update({"height": "2m"})
+            #convert units to Kelvin for compatiabillity with NetCDF-CF conventions
```
```suggestion
            #convert units to Kelvin for compatibility with CMOR MIP table units
```
The netCDF-CF convention just specifies you need to use units from [UDUNITS-2](https://docs.unidata.ucar.edu/udunits). degC is a valid unit there.

However, the units are standardized from the CMIP comparisons in so called "MIP tables". For example, the [CMIP6_day table](https://github.com/PCMDI/cmip6-cmor-tables/blob/main/Tables/CMIP6_day.json).

src/ewatercycle/_forcings/caravan.py
```diff
@@ -209,6 +209,16 @@ def generate(  # type: ignore[override]
...
+            if (ds_basin[temp].attrs["unit"]) == "°C":
+                ds_basin[temp].values = ds_basin[temp].values + 273.15
+                ds_basin[temp].attrs["unit"] = "°K"
```
```suggestion
                ds_basin[temp].attrs["unit"] = "K"
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -209,6 +209,16 @@ def generate(  # type: ignore[override]
...
+        
+        for var in ["evspsblpot", "pr"]:
+            #convert units to kg m-2 s-1 for compatiabillity with NetCDF-CF conventions
+            if (ds_basin[var].attrs["unit"]) == "mm":
+                ds_basin[var].values = ds_basin[temp].values / (86400) #NOTE THAT THIS CONVERSION ASSUMES DAILY DATA
```
BSchilperoort: "All time series data except streamflow are aggregated (daily and spatially over basins)", Kratzert 2023.

And all these other time series data is merged with the streamflow on the same time dimension in the netCDF files we host on 4TU. So I think it's safe to hard code the conversion here.
```suggestion
                ds_basin[var].values = ds_basin[temp].values / (86400)  # mm/day --> kg m-2 s-1
```

BSchilperoort (changes requested):
Having the Caravan forcing homogenized with the other forcings is a good idea. Perhaps we should add some info on this to the documentation so it is easier for people to refer to that (e.g. a table with units and some references).

It's generally not best practice to change things like this "silently" on a non-major release, but I assume that the only other user of the Caravan forcing right now is @Daafip, so he will be aware of the changes.

Can you also update the changelog and perhaps add unit test?

Daafip:
> I assume that the only other user of the Caravan forcing right now is @Daafip, so he will be aware of the change

Yup, for my thesis actually using the original camels dataset still as Caravan is fully Era5 based. Adding Caravan was more of an exploratory extra. 

Don't have time to do any extensive testing but looks good.

Daafip (approved):

BSchilperoort (approved):
I applied the fixes we discussed last week. Tests (should) pass now.

I'll merge this into #420 to have it included in the latest release.

PR #420, opened 2024-05-31
Release prep for 2.1.1

BSchilperoort:
Some bug have been fixed lately, along with other improvements such as the Caravan forcing generator. It would be nice if these were accessible for everyone.

Note that I just turned on the zenodo integration again, for some reason it was turned off again...

Also, I added myself to the zenodo json file, and @Daafip  to that file + CITATION.cff


sverhoeven (approved):
Looks good to me.

Issue #441, opened 2024-07-08
Bump required Python + ruff + no more setup.py|setup.cfg


sverhoeven:
We should do a pass to update everything except the code and test suite.

Must haves
- [x] Python 3.12
- [x] Ruff replaces flake8 + black + isort
  - [x] in lint ghaction run ruff and mypy
- [x] Delete setup.cfg + setup.py + (MANIFEST.in?)
- [x] Drop bump2version
  - use https://github.com/sphinx-toolbox/sphinx-pyproject to get version for docs/conf.py?
- [x] clean precommit, 
   -  drop bad mypy precommit hook
 - [x] drop .zenodo.json use RSD instead of https://zenodo.org/communities/nlesc/records
 - [x] use new pypi publish ghaction as it no longer needs a secret
 - [x] reproducable Lock file for dependencies

Nice to haves
- less config files
  - [x] Move docs/requirements.txt to docs optional dep in pyproject.toml
  - [x] does ruff to editorconfig stuff?
  - [x] .zendo.json is only used for community, replace with https://research-software-directory.org/organisations/netherlands-escience-center?tab=software&order=is_featured
- stop using sonar, 
  - [x]  use codecov for coverage, 
  - [x] ruff will be green or red, you will not have any warnings to show. So we dont have 2 different linters (precommit/flake/ruff and sonarcloud)
  - [x] replace sonarcloud badges with mypy/ruff/coverage badges
- [x] bump docs to latest/best versions
  - [x] update intersphinx urls
  - [x] remove unused napoleon_custom_sections ?
  - [ ] fix docstrings as ` pytest --doctest-modules src` is failing too much
- [x] .vscode settings that installs python, ruff extensions
- [x] .devcontainer that installs deps
- [ ] publish to test pypi on each push

sverhoeven:
We could move .readthedocs.yaml to docs/ and set path at https://readthedocs.org/dashboard/ewatercycle/edit/

BSchilperoort:
> Delete MANIFEST.in?

You can specify files to include/exclude in the packaging with hatch https://hatch.pypa.io/latest/config/build/#patterns

BSchilperoort:
> use https://github.com/sphinx-toolbox/sphinx-pyproject to get version for docs/conf.py?

Adding this should solve that:
```py
from importlib.metadata import version as get_version

# The short X.Y version.
version = get_version("ewatercycle").split("+")[0]
# The full version, including alpha/beta/rc tags.
[...]

Issue #440, opened 2024-07-08
Bump required Python + ruff + no more setup.py


sverhoeven:

sverhoeven:
Dup of #441

PR #443, opened 2024-07-23
Switched to hatch + deleted unneeded setup.cfg and setup.py

sverhoeven:
Also moves mypy from precommit to ci job

Refs #441


pyproject.toml
```diff
@@ -17,3 +98,16 @@ max-line-length = "88"
...
 [tool.pydocstyle]
 convention = "google"
+
+[tool.coverage.run]
+source= "src"
+branch = true
+parallel = true
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
```
BSchilperoort, line 115: src is missing here (for doctest)

pyproject.toml
```diff
@@ -17,3 +98,16 @@ max-line-length = "88"
...
+
+[tool.mypy]
+plugins = ["pydantic.mypy"]
+ignore_missing_imports = true
+files = ["src", "tests"]
```
BSchilperoort: I'm not a big fan of checking the typing of the tests, can be a hassle.

Also, we can add `check_untyped_defs = True` (but that might give some new mypy errors)

`warn_unused_ignores = True` would be good to add.

BSchilperoort (approved):
Looks good, only some small comments.

The built tarball includes many files currently, and MANIFEST.in is being ignored (can be deleted). You can use excludes to reduce the tarball size.

pyproject.toml
sverhoeven: Skipped as most docstrings have unmet expectations, added as nice to have to #440

pyproject.toml
sverhoeven: Added both

sverhoeven:
Thanks for checking and your suggestions.

BSchilperoort:
> Thanks for checking and your suggestions.

No problem.

By the way, MANIFEST.in is still there, but is no longer doing anything

sverhoeven:
> > Thanks for checking and your suggestions.
> 
> No problem.
> 
> By the way, MANIFEST.in is still there, but is no longer doing anything

Made https://github.com/eWaterCycle/ewatercycle/pull/444 to remove it.

PR #444, opened 2024-07-24
No longer need MANIFEST.in

sverhoeven:
Remove remnant of old build system

Refs #441


BSchilperoort (approved):
Fewer files better :+1:

PR #445, opened 2024-07-24
Replace bumpversion with single version in pyproject.toml

sverhoeven:
Refs #441 

Used suggestions in #441 for runtime.
For docs parse toml file.

To check
- [x] runtime
- [x] local doc build
- [x] ~readthedocs build~ Cannot check version on readthedocs as https://ewatercycle--445.org.readthedocs.build/en/445/ uses pr as version.


BSchilperoort (approved):
Looks good! :rocket:

PR #448, opened 2024-07-29
Use trusted publishing

sverhoeven:
Refs #441 

On pypi added this ci workflow as a trusted published.


BSchilperoort (approved):
Looks good. Would be more future-proof than pypi account-linked tokens.

It is also possible to create this action for https://test.pypi.org/.  Although that's probably not really needed if building the package is already tested elsewhere (i.e. as in the `ruff` branch).

sverhoeven:
Thanks for reviewing. 

After a release is made and the workflow works, the PYPI_API_TOKEN secret can be removed.

PR #449, opened 2024-07-31
Replace sonarcloud with codecov and lint ci badge

sverhoeven:
Refs #441 

Also renames the ci workflows to match their content.


BSchilperoort (approved):
Codecov seems to work, nice :+1: 

The one thing I dislike from an aesthetic view is how github names the actions:
image

Perhaps you can style them like this:
```
Build distribution 📦 / build
Static code analysis 🔍️ / lint
Testing and type checking 🐛 / test
```
It would then be a bit more clear which workflow does what. Of course the emojis are optional haha

PR #454, opened 2024-08-29
Remove .readthedocs.yaml

sverhoeven:


BSchilperoort (approved):
I also missed that the original file wasn't removed

PR #455, opened 2024-08-29
Prep for release 2.3.0

### LINKED ISSUE #304: add curl example to readme.md install section
As per the request of the reviewers and editors of the GMD paper, add a (tested) curl based version of the install instructions.

### LINKED ISSUE #357: Using conda-lock to avoid dependencies breaking things
Lately many issues have arisen related to changes elsewhere breaking the eWaterCycle platform. E.g. #355. This issues can arise from the dependencies of our dependencies (i.e. ESMValTool's many dependencies), so pinning our direct dependencies is insufficient.

With [conda-lock](https://github.com/conda/conda-lock) we could create a completely reproducible environment, where all (sub)dependencies are pinned. This way we will not be surprised by things breaking suddenly.

As a bonus conda installs will be much faster as the conda package resolver is not needed anymore.

One downside is that we will manually have to upgrade the lock file. Sadly Dependabot does not support conda environments.

### LINKED ISSUE #421: Update readme example
The Marrmot14 code in the example won't run, it's not up to date with 2.0+

sverhoeven:
Fixes #421 
Fixes #304


BSchilperoort (approved):
Nice!

Closes #357

PR #458, opened 2024-09-17
Support the newest OPeNDAP version of Caravan, patch data issues.

BSchilperoort:
This PR fixes some issues in the already hosted dataset:

- homogenizes variable names
- converts all units
- massages the data into the right format to have the data the users needed

It also adds a new kwarg to download the shapefiles, by default downloading the shapefile is now skipped.


BSchilperoort:
@RolfHut could you try this branch to see if it fits with what you started in #433

PR #464, opened 2024-11-07
Add get_shape_area() and plot_shape() methods to DefaultForcing

RolfHut:
first commit added get_shape_area (needs testing)

### DIFF
--- a/src/ewatercycle/base/forcing.py
+++ b/src/ewatercycle/base/forcing.py
@@ -39,6 +39,11 @@ import fiona
 import shapely
 from pyproj import Geod
 
+import shapely.geometry
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import cartopy.feature as cfeature
+
 import xarray as xr
 from pydantic import BaseModel
 from pydantic.functional_validators import AfterValidator, model_validator
@@ -306,7 +311,7 @@ class DefaultForcing(BaseModel):
     def get_shape_area(self) -> float:
         """Return the area of the shapefile in m2.
         """
-        if shape == None:
+        if forcing.shape == None:
             raise ("Shapefile not specified")
         shape = fiona.open(forcing.shape)
         poly = [shapely.geometry.shape(p["geometry"]) for p in shape][0]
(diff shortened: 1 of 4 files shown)


RolfHut:
tested and works, but check fails, I could use some tips on the checks that are being run and the best practices we adhere to in coding style

BSchilperoort:
Thanks for picking this up! There are instructions in the [CONTRIBUTING.md document](https://github.com/eWaterCycle/ewatercycle/blob/main/CONTRIBUTING.md#you-want-to-make-some-kind-of-change-to-the-code-base).

I think the instructions are missing that `pip3 install -e .` should actually be `pip3 install -e .[dev]` (so the dev dependencies are installed as well).

To run the pre-commit hook manually (on all files, not associated with a commit), you can do `pre-commit run --all`. This should fix the formatting issues.

The codecov check is failing because you've added code that is not tested. If you would add tests for the two functions this should be fixed. For the shape area you can easily check if the output is numerically correct. Testing the plotting code is more difficult, maybe we only check if it ran without errors.

src/ewatercycle/base/forcing.py
```diff
@@ -299,6 +309,50 @@ def variables(self) -> tuple[str, ...]:
...
+
+        shape = fiona.open(self.shape)
+        poly = [shapely.geometry.shape(p["geometry"]) for p in shape][0]
+        w, s, e, n = poly.bounds  # different order than set_extent expects
+        
+        # 10 % of the minimum of either the west-east extend or the north-south extend is used as
+        # padding around the shape.
+        pad = min(abs(w - e), abs(n - s)) * 0.1 
+        
+        plt.figure(figsize=(8, 5))
```
BSchilperoort: It would be nicer to allow a user to provide a matplotlib axis to plot the shape in (optionally), and make a new figure if no axis is provided.

src/ewatercycle/base/forcing.py
```diff
@@ -299,6 +309,50 @@ def variables(self) -> tuple[str, ...]:
         """
         return tuple(self.filenames)
 
+    def get_shape_area(self) -> float:
+        """Return the area of the shapefile in m2.
+        """
+        if self.shape == None:
+            raise ("Shapefile not specified")
```
BSchilperoort:
```suggestion
            raise ValueError("Shapefile not specified")
```

src/ewatercycle/base/forcing.py
```diff
@@ -299,6 +309,50 @@ def variables(self) -> tuple[str, ...]:
...
+    def plot_shape(self) -> matplotlib.axes._axes.Axes:
+        """Make a plot of the shapefile on a map of the world.
+        The plot is padded with 10% of the north-south extend,
+        or the east-west extend (which ever is smallest).
+
+        the axis object is ter
+        """
+
+        if self.shape == None:
+            raise ("Shapefile not specified")
```
BSchilperoort:
```suggestion
            raise ValueError("Shapefile not specified")
```

src/ewatercycle/base/forcing.py
```diff
@@ -299,6 +309,50 @@ def variables(self) -> tuple[str, ...]:
...
+        plt.figure(figsize=(8, 5))
+        ax = plt.axes(projection=ccrs.PlateCarree())
+        ax.add_geometries(
+            poly, crs=ccrs.PlateCarree(), facecolor='#f5b41d', edgecolor='k', alpha=0.8,
+        )
+        ax.add_feature(cfeature.COASTLINE, linewidth=1)
+        ax.add_feature(cfeature.RIVERS, linewidth=1)
+        ax.add_feature(cfeature.OCEAN, edgecolor="none", facecolor="#4287f5")
+        
+        ax.set_extent((w-pad, e+pad, s-pad, n+pad), crs=ccrs.PlateCarree())
```
BSchilperoort:
```suggestion
        ax = plt.axes(projection=ccrs.PlateCarree())
        ax.add_geometries(  # type: ignore
            poly, crs=ccrs.PlateCarree(), facecolor='#f5b41d', edgecolor='k', alpha=0.8,
        )
        ax.add_feature(cfeature.COASTLINE, linewidth=1)  # type: ignore
        ax.add_feature(cfeature.RIVERS, linewidth=1)  # type: ignore
        ax.add_feature(cfeature.OCEAN, edgecolor="none", facecolor="#4287f5")  # type: ignore
        
        ax.set_extent((w-pad, e+pad, s-pad, n+pad), crs=ccrs.PlateCarree())  # type: ignore
```
Typing doesn't work well with matplotlib... This ignores the typing on the offending lines, and should make the tests

BSchilperoort (commented):
These suggestions should fix most issues.

BSchilperoort:
I've fixed the linting/formatting issues and added tests for the new methods. Everything's good now :heavy_check_mark:

BSchilperoort (approved):

Issue #231, opened 2021-08-09
Add method or log message to find the model log file


Peter9192:
When you start a model (e.g. pcrglobwb), it can take some time where it might seem like the kernel is hanging. In the meantime, model log messages are printed in the model working directory. It might be useful to add a console message like 
`f"Initializing model. Log messages will be written to {cfg_dir}/default.log` or something like that.

Can this be added as a (model-specific) method to the model class?

sverhoeven:
Some models only use stdout and stderr for logs. They are captured by default and can be viewed with `self.bmi.logs()`. See `capture_logs` argument on https://grpc4bmi.readthedocs.io/en/0.2.x/api/grpc4bmi.bmi_client_singularity.html

So depending of flavour of model we could print 
- Log messages will be written to {cfg_dir}/default.log
- Log messages can be requested with model.bmi.logs()

PR #470, opened 2024-12-10
Give users easier access to container logs

### LINKED ISSUE #469: Wrapping of BMI models with OptionalDestBmi/MemoizedBmi drops .logs()
Intended behavior is that you can do: `model.bmi.logs()` and see the container's output (https://github.com/eWaterCycle/ewatercycle/issues/231)

However, during the wrapping of the model with the OptionalDestBmi and MemoizedBmi this class method is buried;
https://github.com/eWaterCycle/ewatercycle/blob/3caa85ebcad22ffda40ee08e31fe646c4574883c/src/ewatercycle/container.py#L211-L213

It is still accessible with; `model.bmi.origin.origin.logs()` however that's not ideal.

We could implement a `model.logs` property that will return the logs, by digging them up again through the layers.

BSchilperoort:
Closes #469, #231 

Usage example:
image


BSchilperoort:
@RolfHut to review you could just test the functionality in a notebook you are already using somewhere else.

Install this in-dev version of ewatercycle in your terminal with:
```sh
python -m pip install --no-deps https://github.com/eWaterCycle/ewatercycle/archive/refs/heads/main.zip
```
This should not mess up your environment (due to the `no-deps` flag only this package will be updated).

<hr>

I know that the code coverage check is failing, but it's not trivial to write a test for this case and time is sadly limited.

BSchilperoort:
@ahmedfarhatx22 this should give you easy access to the model container logs :)

RolfHut:
machines already running do not have remote BMI and now fall over...

RolfHut:
in testing on HBV I get this error:  

```'HBV' object has no attribute 'logs'```

screenshot:
Screenshot 2024-12-11 at 17 50 17

 so maybe something goed wrong in error handling if a model does not support this?

**Comment by BSchilperoort** [comment]
> in testing on HBV I get this error:
> 
> `'HBV' object has no attribute 'logs'`
> 
> screenshot: Screenshot 2024-12-11 at 17 50 17
> 
> so maybe something goed wrong in error handling if a model does not support this?

It's a property of the ContainerizedModel class: any (grpc) containerized model should work.

I did add a check if the model is set up yet in 388f748, and improved the error messages.

**Comment by RolfHut** [comment]
Screenshot 2024-12-12 at 09 07 27

also a no at leaky bucket

**Comment by RolfHut** [comment]
> > in testing on HBV I get this error:
> > `'HBV' object has no attribute 'logs'`
> > screenshot: Screenshot 2024-12-11 at 17 50 17
> > so maybe something goed wrong in error handling if a model does not support this?
> 
> It's a property of the ContainerizedModel class: any (grpc) containerized model should work.
> 
> I did add a check if the model is set up yet in [388f748](https://github.com/eWaterCycle/ewatercycle/commit/388f7485b35aba1513848cc2c7b1a41ad0c83257), and improved the error messages.

how do I pull your commit on my SRC machine? than I can check if this fixes it.

**Comment by BSchilperoort** [comment]
I realize now I gave you the wrong command, sorry.

It should have been
`python -m pip install --no-deps https://github.com/eWaterCycle/ewatercycle/archive/refs/heads/access-logs.zip`

the other one was for `main`

**Comment by RolfHut** [comment]
tried that and same error on leaky bucket: no attribute 'logs'

File: src/ewatercycle/base/model.py [hunk 1]
```diff
@@ -452,6 +453,27 @@ def version(self) -> str:
...
+    def logs(self) -> str:
+        """Return the container logs."""
+        if not hasattr(self, "_bmi"):
+            msg = (
+                "The model has no BMI attached (yet).\n"
+                "Logs will only be available after running `.setup`"
+            )
+            raise ValueError(msg)
+        bmi = self._bmi
+        while isinstance(bmi, OptionalDestBmi | MemoizedBmi):
```
BSchilperoort: This won't work yet for the pcrglobwb plugin due to [this](https://github.com/eWaterCycle/ewatercycle-pcrglobwb/blob/7487c6198e0f03ba8e7a9962510348655142eed5/src/ewatercycle_pcrglobwb/model.py#L23)

the SwapXY class uses https://github.com/eWaterCycle/ewatercycle/blob/main/src/ewatercycle/container.py#L342

sverhoeven (approved):
Ran the notebook with model_instance.logs calls successfully.

However it would be nice to have some unit tests that take less work to setup.

BSchilperoort:
> However it would be nice to have some unit tests that take less work to setup.

I considered this, but the logs are part of the containerized model, where testing relies on mocking a lot. As we don't actually start a container, I would have to mock the starting of the container, and mock the log return. At that point not much would be left to actually test as it's only mocks. The classes which actually implement the logs are in grpc4bmi and remotebmi. It would be good for this package to actually have a containerized model integrated which is used in unit tests, but that's not in the scope of this PR.

Issue #473, opened 2025-02-07
Broken link in README.md


BSchilperoort:
In the top discription;

> Exposes [simple interface](https://ewatercycle.readthedocs.io/en/latest/user_guide.html) to quickly get up and running

points to the old single-html user guide. This was split up, so the URL should refer to https://ewatercycle.readthedocs.io/en/latest/user_guide/00_intro.html

BSchilperoort:
@MarkMelotto

PR #474, opened 2025-02-07
473 fix readme

MarkMelotto:
Fixed issue #473, the hyperlink now points to the correct link.
@BSchilperoort  could you check this :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,9 @@ models.
 - Supports observation data from [GRDC or
     USGS](https://ewatercycle.readthedocs.io/en/latest/observations.html)
 - Exposes [simple
-    interface](https://ewatercycle.readthedocs.io/en/latest/user_guide.html)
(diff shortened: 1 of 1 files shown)


BSchilperoort (approved):
That's indeed the required fix! First PR on ewatercycle :partying_face: 

Usually we would add one or more entries in the changelog as well when merging a PR, but I think this is so minor that it's not needed this time.

Do remember to add yourself to the CITATION.cff file in the next PR. This way you will show up on Zenodo etc. as a co-author of this package.

The linter in the CI seems to fail, however that's probably due to a change in ruff's code formatting rules. That can be fixed in a different PR later.

When you merge the PR make sure to do `Squash and merge`. This way the commits are collapsed into a single commit, and a link to this PR is made for future reference (so any discussions in the PR can be found back at a later point in time).

PR #479, opened 2025-05-20
478 caravan forcing downloads and unzips shapefiles for all caravan with each basin leads to errors fixes #478

### LINKED ISSUE #478: caravan forcing downloads and unzips shapefiles for all caravan with each basin: leads to errors
When generating forcing from caravan, the entire caravan shapefile (100 MB) is downloaded and unzipped each time, unless a path to the shapefile of the individual basin is provided. Timing issues between python and OS (I think?) cause this to fail when multiple catchments are requested in rapid succession. 

As a researcher who wants to study the entire camels dataset for my research, I want to be able to point to a location where the combined shapefile is stored (cached) to avoid this overhead.

A solution would be to check if the current optional "shape" argument points to a file called "combined.shp". If it does, it is assumed this is the combined caravan shapefile and the shape of the selected catchment is extracted from it.

RolfHut:
allowed the option in shape to point to combined.shp file that is already downloaded. Removed the check on polygons, which threw an error that was not correct stopping workflows

### DIFF
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -188,7 +188,6 @@ class CaravanForcing(DefaultForcing):
         ds_basin = ds.sel(basin_id=basin_id.encode())
         ds_basin_time = crop_ds(ds_basin, start_time, end_time)
 
-        
         if shape_in is None:
             shape = get_shapefiles(Path(directory), basin_id)
         elif Path(shape_in).name == "combined.shp":
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -164,7 +164,8 @@ class CaravanForcing(DefaultForcing):
             directory: Directory in which forcing should be written.
             variables: Variables which are needed for model,
                 if not specified will default to all.
-            shape_in: (Optional) Path to a shape file of the basin, or the combined.shp file of all basins.
+            shape_in: (Optional) Path to a shape file of the basin, or the combined.shp 
+                file of all basins.
                 If none is specified, will be downloaded automatically.
             kwargs: Additional keyword arguments.
                 basin_id: The ID of the desired basin. Data sets can be explored using
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -164,7 +164,7 @@ class CaravanForcing(DefaultForcing):
             directory: Directory in which forcing should be written.
             variables: Variables which are needed for model,
                 if not specified will default to all.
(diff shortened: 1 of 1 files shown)


MarkMelotto (changes requested):
Do we want to make some tests for this?
As it should be done and can easily be implemented!

RolfHut:
we absolutely want to write some tests for this.

I think we should have at least one model run test (HBVLocal?)

MarkMelotto:
@RolfHut we can do that.
Or just basic known shapefiles and other things that are not exactly shapefiles ;)

src/ewatercycle/_forcings/caravan.py
```diff
@@ -188,8 +189,19 @@ def generate(  # type: ignore[override]
         ds_basin = ds.sel(basin_id=basin_id.encode())
         ds_basin_time = crop_ds(ds_basin, start_time, end_time)
 
-        if shape is None:
+        if shape_in is None:
             shape = get_shapefiles(Path(directory), basin_id)
+        elif Path(shape_in).name == "combined.shp":
```
BSchilperoort, line 194: This would not be the most robust way of checking for an already existing combined shapefile. It's also a bit difficult to document.

Maybe it would be better to;

- Not download automatically anymore if `shape=None`, although this breaks existing code using this feature
- Add a keyword argument (like the existing `basin_id` one) for downloading the shapes, `download_shape` which defaults to false.
```py
# get download_shape from kwargs, default to false;
download_shape = kwargs.get("download_shape", False) 
```
- Add another kwarg  `combined_shapefile` that you can use to point to the combined shapefile with all basins
```py
combined_shape = kwargs.get("combined_shape")
if combined_shape is not None:
    combined_shape = Path(combined_shape)
    if not combined_shape.exists():
        msg = "Could not find combined shapefile."
        raise FileNotFoundError(msg)
```

These kwargs can be properly documented in the docstring, and validating the user input is a bit easier this way.

BSchilperoort:
> Do we want to make some tests for this?

Just be aware that downloading data from the internet during testing is not very desirable. You can have some test data included in the repository however, and "mock" the download of the test so it just points to that file. See https://docs.pytest.org/en/stable/how-to/monkeypatch.html

> I think we should have at least one model run test (HBVLocal?)

I think this would be good, but the tests should also be able to run without having a model installed.
You can do this by only running these tests if you can import the package;

```py
def model_available()
    try:
        import ewatercycle_leakybucket
        return True
    except ModuleNotFoundError:
        return False

MODEL_AVAILABLE = model_available()

@pytest.mark.skipif(not MODEL_AVAILABLE, reason="requires leakybucket to be installed")
def test_function(): ...
```
See https://docs.pytest.org/en/stable/how-to/skipping.html#id1

Issue #480, opened 2025-09-05
(minor) bug report: Incorrect line numbers in GRDC metadata reader


andrevdv:
In `def _grdc_metadata_reader` the line numbers used to extract metadata from .txt are off by one:

https://github.com/eWaterCycle/ewatercycle/blob/ce0099673757ece2f3026c841290312ec9323185/src/ewatercycle/observation/grdc.py#L354-L361

- `dataSetContent` is currently read from index 20, but should be 21 to read line 22 in the .txt
- `units` is currently read from index 22, but should be 23 to read line 24 in the .txt

Additionally, the "units" attribute is not currently used, units are currently hardcoded in the description when creating the xarray object

Proposed fix:

- Update the indices to 21 and 23.
- Optionally remove or repurpose the unused "units" attribute to avoid confusion.

PS: As mentioned to @RolfHut, I’m currently working on extending the GRDC station reader to also handle stations with monthly data instead of daily. I could bundle these bugfixes with that, if preferred.

<img width="702" height="184" alt="Image" src="https://github.com/user-attachments/assets/087a1f8c-5082-4230-b4b7-90cdcacf1fab" />

MarkMelotto:
@andrevdv thank you very much :)
I have fixed the bug and await the approval of @RolfHut 

for the "units" part, if you can fix this together with what you are working on that would be amazing!

PR #481, opened 2025-09-08
small bug fix

MarkMelotto:
fix for issue #480


MarkMelotto:
@RolfHut finally fixed the linter now it is ready :)

src/ewatercycle/base/forcing.py
```diff
@@ -388,6 +388,27 @@ def __getitem__(self, key: str) -> Path:
...
+        The hash is computed using the object's start_time, end_time, directory,
+        shape, and filenames attributes. This allows Forcing instances to be used
+        in hashed collections such as sets and as dictionary keys, provided these
+        attributes are themselves immutable and hashable.
+
+        Returns:
+            int: The hash value of the object.
+        """
+        return hash(
+            (
```
BSchilperoort: For a proper hash you are missing some info here. E.g. the dataset definition. So `str(self.dataset)`should also be included. Some forcing implementations add more kwargs as well, making this even more complex to maintain.

Might be better to remove the linter warning and not implement `__hash__` for this object.

src/ewatercycle/observation/grdc.py
```diff
@@ -351,12 +351,12 @@ def _grdc_metadata_reader(grdc_station_path, all_lines):  # noqa: C901
             attribute_grdc["altitude_masl"] = "NA"
 
         try:
-            attribute_grdc["dataSetContent"] = str(all_lines[20].split(":")[1].strip())
+            attribute_grdc["dataSetContent"] = str(all_lines[21].split(":")[1].strip())
```
BSchilperoort: Would be nicer to extract this based on the actual header info/line start.
The line with this info seems to start with "# Data Set Content" which you should be able to find. Same for the other attributes.

src/ewatercycle/observation/grdc.py
BSchilperoort, line 1: Would be nice to add a test for this module. You can take a real GRDC file, and replace any real values with dummy values to avoid copyright/license issues.

MarkMelotto:
@BSchilperoort is there a way to tell the linter that the __hash__ can be ignored?

BSchilperoort:
> @BSchilperoort is there a way to tell the linter that the **hash** can be ignored?

https://docs.astral.sh/ruff/linter/#error-suppression

MarkMelotto:
@BSchilperoort the final format does not work for me locally.
When I run the format, it says all files already formatted...

So what is the next plan of action then?

BSchilperoort:
Also couldn't reproduce this locally. I've enabled the --diff flag to debug the ruff formatter. Make sure you git pull before continuing on this branch

BSchilperoort:
Not sure why ruff on the CI wants this, but at least you can solve it this way;

```diff
--- src/ewatercycle/util.py
+++ src/ewatercycle/util.py
@@ -225,9 +225,9 @@
         #   https://github.com/eWaterCycle/infra/issues/157
         #   the following is a workaround.
         if "time_bnds" in datasets[i] and xr.infer_freq(datasets[i]["time"]) == "D":
-            datasets[i]["time"] = datasets[i]["time_bnds"].isel(
-                bnds=0
-            ) + pd.Timedelta("12H")
+            datasets[i]["time"] = datasets[i]["time_bnds"].isel(bnds=0) + pd.Timedelta(
+                "12H"
+            )
             datasets[i] = datasets[i].drop_vars("time_bnds")
 
         # A "height" coordinate can be present, which will result in conflicts.
```

Hm. That line looks like this to me locally;
```py
            datasets[i]["time"] = datasets[i]["time_bnds"].isel(
                bnds=0
            ) + np.timedelta64(12, "h")
```
which is not the same

BSchilperoort:
Ah makes sense now. Do `git checkout main; git pull; git checkout grdc-patch; git merge main`.
CI is run on `main` with this PR applied. Main has new commits

MarkMelotto:
yep got it, thanks for the help!

MarkMelotto:
@BSchilperoort I can still write some test functions for the new function for grdc, other than that. I think it is good now!

src/ewatercycle/observation/grdc.py
```diff
@@ -300,71 +300,50 @@ def _grdc_metadata_reader(grdc_station_path, all_lines):  # noqa: C901
...
+    """Private helper to extract metadata fields from GRDC header lines."""
+    for line in lines:
+        if line.strip().startswith(f"# {key}:"):
+            parts = line.split(":", 1)
+            if len(parts) > 1:
+                value = parts[1].strip()
+                try:
+                    return cast(value)
+                except ValueError:
+                    return default
```
BSchilperoort, line 352: I think this can lead to unexpected behavior. Might be better to raise an error with a clear message here (or use python's default error if that one is clear).

BSchilperoort (commented):
Just one comment on the try/except block. Otherwise it looks good.

MarkMelotto:
@BSchilperoort done

src/ewatercycle/observation/grdc.py
BSchilperoort, line 352: If you return the default when _casting_  failed, it would be better to warn the user. Upside is that there's no error and the rest of the data loads fine, but then the user is also aware of a line that wasn't read properly.

MarkMelotto: yep you are right, yesterday wasnt the best day I guess :)

MarkMelotto:
@BSchilperoort this is better now :)

BSchilperoort (approved):
looks good now

PR #485, opened 2025-10-01
Add shapefile plotting utility function

### LINKED ISSUE #484: Plot shapefile/catchment utility function
For demo notebooks or in a student thesis it can be nice to show the catchment that is being modeled on a map. This would reduce the clutter and can implement a consistent and clean API to plot.

Something like;
```py
def plot_catchment(
    shape: Path | str,
    lat_bounds: tuple[float, float] | None,
    lon_bounds: tuple[float, float] | None,
    ax: matplotlib.axis.Axis | None,
) -> tuple[matplotlib.figure.Figure, matplotlib.axis.Axis] | None:
    # Load shape / geometry

    if ax:
        # plot in existing axis
    else:
        # create cartopy mat plot, add shapefile
        # also add country borders, coastline, sea, rivers.

    if lat_bounds and lon_bounds:
        # apply bounds (set_xlim, set_ylim)
    else:
        # infer appropriate bounds, apply

    if ax is None:
        return fig, axis # return figure and axis so users can add to/modify plot
```

BSchilperoort:
Closes #484 

Would still be nice to add to user docs somewhere.


BSchilperoort:
image

plot will look like this

src/ewatercycle/util.py
```diff
@@ -353,3 +359,94 @@ def get_package_versions() -> dict[str, str]:
...
+            lon_bounds = (minx - xpad, maxx + xpad)
+        if lat_bounds is None and axis is None:
+            lat_bounds = (miny - ypad, maxy + ypad)
+
+    if lon_bounds is not None:
+        ax.set_xlim(lon_bounds)
+    if lat_bounds is not None:
+        ax.set_ylim(lat_bounds)
+
+    if axis is None:
```
MarkMelotto, line 451: So you return none or an empty figure?

MarkMelotto (approved):

src/ewatercycle/util.py
BSchilperoort: if the user provides an axis I return nothing (user already has fig, ax). Only if user does not provide a plot axis, I generate one and give the figure and axis handle to the user (e.g. for plotting in more data, modifying styling)

MarkMelotto: aah of course, smart

Issue #491, opened 2025-12-08
4tu has issues, so we want to use dcache for that


MarkMelotto:

MarkMelotto:
opened a branch for this: caravan-dcache

BSchilperoort:
Could you elaborate on what's wrong with 4tu? :)

MarkMelotto:
@BSchilperoort unreliable, we keep getting openDAP or netCDF errors.
Once in 10 times it does work, they are working on it in parallel, but it has been some time already.

BSchilperoort:
It would be nicer if they would host Zarr stores as well on 4TU. Should be much simpler...

MarkMelotto:
Should be fixed with PR #492

PR #492, opened 2025-12-08
add fallback data to caravan

MarkMelotto:
Hi @BSchilperoort 

Fixing issue #491 .

Since I am not that familiar with coding style of ewatercycle, I require your help.
The problem we are facing is that caravan on 4tu is not working reliably at the moment, so I made this fallback version myself.

The crop_ds change is now needed, because we now have date instead of time, need to find another fix I think, let me know what you think!

And please help me for the overall picture as to what i can approve.

Thanks!


MarkMelotto:
before writing tests, I would like the feedback of Bart

BSchilperoort:
> before writing tests, I would like the feedback of Bart

It might be simpler to define an environmental variable `CARAVAN_CACHE` (containing the directory of the caravan file), and have xarray open that if it's available;

```py
import os

if os.environ.get("CARAVAN_CACHE"):
    # load from cache
else:
    # load from 4TU
```

This would be a minimal change in the code, and would remove the need for users to pass things.

Just make sure you download the caravan file from the HTTP link; https://opendap.4tu.nl/thredds/fileServer/data2/djht/ca13056c-c347-4a27-b320-930c2a4dd207/3/camels.nc so the file is identical to the OPeNDAP version.

MarkMelotto:
thanks for the insights!
The HTTP should still be the same right?

For now when I am working on Snellius I have caravan data mounted to something other then...
Oh wait I get it, I need to do some more reading into environmental variables!

Also, 1.3 GB that is not much :)

MarkMelotto:
I am also noticing how you use /3/ in your opendap url and not /1/ as it is in the code.
I put all the /1/ datasets on dcache now, so I just have to figure out how to set the environmental variable stuff.

Daafip:
@MarkMelotto the /3 indicates v3, we did some finetuning. Nice to see you're using the data!

MarkMelotto:
hey @Daafip !

I noticed that v3 only contains the camels.nc, not the other ones, is that correct?
So I pulled them from v1.

Please let me know!

MarkMelotto:
@BSchilperoort so now I would need to add:
export CARAVAN_CACHE=~/data/shared/climate-data/caravan/

to a notebook in SRC to have this work if I understand correctly?

src/ewatercycle/_forcings/caravan.py
```diff
@@ -118,6 +119,11 @@ def get_dataset(cls: type["CaravanForcing"], dataset: str) -> xr.Dataset:
                 'hysets',
                 'lamah'
         """
+        cache_dir = os.environ.get("CARAVAN_CACHE")
+        # Check if we want to load from 4TU or dCache
+        if cache_dir:
+            cache_dir = cache_dir.rstrip("/")  # ensure no trailing slash issues
```
BSchilperoort: or do 
```py
xr.open_dataset(Path(cache_dir) / "{dataset}.nc")
```

BSchilperoort:
> export CARAVAN_CACHE=~/data/shared/climate-data/caravan/

or add that to the end of your `~/.bashrc` file. That will make it always available _for you_.

MarkMelotto:
@BSchilperoort  alright thanks for the help.
Then I guess I should write a test and then we can push to main?

Then for new SRC machines the correct version will be installed already, but for the other we need to update them.

BSchilperoort:
> Then I guess I should write a test and then we can push to main?
> 
> Then for new SRC machines the correct version will be installed already, but for the other we need to update them.

I think you also need to make a new release, and update infra for it to be on new SRC machines.

Note that for SRC you should be able to define environment vars for all users as well. For this you will have to modify the infra repo.

See;
- https://github.com/eWaterCycle/infra/blob/c02600be0b135577da9682d5b4986b81c8162ac5/roles/jupyter/tasks/main.yml#L75-L79
- https://github.com/eWaterCycle/infra/blob/c02600be0b135577da9682d5b4986b81c8162ac5/roles/jupyter/templates/profile.sh.j2
- https://github.com/eWaterCycle/infra/blob/c02600be0b135577da9682d5b4986b81c8162ac5/roles/jupyter/README.md#role-variables
- background info; https://stackoverflow.com/a/1641531

i.e. seems like modifying the `jupyterhub_spawner_environment` var of the to `jupyter` ansible task, by adding the new environment variable should do the trick. I _think_ the default values for these are defined [here](https://github.com/eWaterCycle/infra/blob/c02600be0b135577da9682d5b4986b81c8162ac5/roles/jupyter/defaults/main.yml#L7)

Should be a good first issue to fix in the ansible setup so you can learn how that works :wink: You should follow Stefan's guide on how to test this locally; https://github.com/eWaterCycle/infra/blob/c02600be0b135577da9682d5b4986b81c8162ac5/VAGRANT.md

MarkMelotto:
phew thanks @BSchilperoort that is a lot of new stuff :D

I think the idea was to have this as a fail safe and still use 4tu as default.
But when needed we provide an opportunity to override the 4tu bit.

So I guess that just leaves the release?

BSchilperoort:
> I think the idea was to have this as a fail safe and still use 4tu as default.

OK then this is fine, yes. You can now override the 4TU default if required. Remember to also update the docstrings to reflect this new behavior.

MarkMelotto:
Hopefully getting the shapefiles will not cause issues...

BSchilperoort:
> Hopefully getting the shapefiles will not cause issues...

You could also use the cache for that. Should a straightforward addition

Daafip:
> I noticed that v3 only contains the camels.nc, not the other ones, is that correct?

v3 is updated 2025-09-16T06:19:23.356Z: 
https://opendap.4tu.nl/thredds/catalog/data2/djht/ca13056c-c347-4a27-b320-930c2a4dd207/3/catalog.html. We updated the structure of the netcdf I believe. 

I think I had a miscommunication with 4TU, the other files from v2 for the other netcdf files are the ones to use. 
https://opendap.4tu.nl/thredds/catalog/data2/djht/ca13056c-c347-4a27-b320-930c2a4dd207/2/catalog.html

MarkMelotto:
> I think I had a miscommunication with 4TU, the other files from v2 for the other netcdf files are the ones to use. https://opendap.4tu.nl/thredds/catalog/data2/djht/ca13056c-c347-4a27-b320-930c2a4dd207/2/catalog.html

Hmm that might be the issue we are facing with 4TU then?
I will also change that!

Daafip:
>  might be the issue we are facing with 4TU then?

I don't thinks so, I cant quickly find the issue why we did the update but there was a reason for it..

src/ewatercycle/_forcings/caravan.py
```diff
@@ -106,7 +107,12 @@ class CaravanForcing(DefaultForcing):
 
     @classmethod
     def get_dataset(cls: type["CaravanForcing"], dataset: str) -> xr.Dataset:
-        """Opens specified dataset from data.4tu.nl OPeNDAP server.
+        """Opens specified dataset from data.4tu.nl OPeNDAP server/specified path.
```
```suggestion
        """Opens specified dataset from data.4tu.nl OPeNDAP server, or cache if available.
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -106,7 +107,12 @@ class CaravanForcing(DefaultForcing):
...
+
+        By default, it will open the dataset from data.4tu.nl OPeNDAP server
+        This can be overridden by having an environmental variable: CARAVAN_CACHE.
+        Pointing to /path/to/camels.nc etc.
```
```suggestion
        This variable should be set to the path of the directory containing the netCDF files.
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -246,7 +257,18 @@ def generate(  # type: ignore[override]
 
 
 def get_shapefiles(directory: Path, basin_id: str) -> Path:
-    """Retrieve shapefiles from data 4TU.nl ."""
+    """Retrieve shapefiles from data 4TU.nl/specified path."""
```
```suggestion
    """Retrieve shapefiles from data 4TU.nl or cache."""
```

src/ewatercycle/_forcings/caravan.py
```diff
@@ -246,7 +257,18 @@ def generate(  # type: ignore[override]
...
+    cache_dir = os.environ.get("CARAVAN_CACHE")
+    # Check if we want to load from 4TU or dCache
+    if cache_dir:
+        shape_path = directory / f"{basin_id}.shp"
+        combined_shapefile_path = Path(f"{cache_dir}/shapefiles/combined.shp")
```
```suggestion
        combined_shapefile_path = Path(cache_dir}) / "shapefiles" / "combined.shp"
```

BSchilperoort (changes requested):
some suggestions

MarkMelotto:
> some suggestions

thank you!
I am struggling with the shapefile pathing...

MarkMelotto:
@BSchilperoort happy new year ;)
Is this now up to scratch?

src/ewatercycle/_forcings/caravan.py
```diff
@@ -118,6 +124,11 @@ def get_dataset(cls: type["CaravanForcing"], dataset: str) -> xr.Dataset:
                 'hysets',
                 'lamah'
         """
+        cache_dir = os.environ.get("CARAVAN_CACHE")
+        # Check if we want to load from 4TU or dCache
+        if cache_dir:
+            cache_dir = cache_dir.rstrip("/")  # ensure no trailing slash issues
```
```suggestion
```
shouldn't be needed anymore because you use `pathlib.Path` on the next line

BSchilperoort (approved):
Happy new year!

Looks good, just one small comment. Nice work!

MarkMelotto:
keeping the branch for a while, in the case of issues arising.

PR #494, opened 2025-12-11
Add support for reading monthly GRDC discharge data (.txt)

andrevdv:
Add support for reading monthly GRDC data from `<station_id>_Q_Month.txt`.  

Unlike daily data, monthly files include three columns:  
- `Original` discharge  
- `Calculated` discharge 
- `Flag` indicating the percentage of valid daily values used for the calculation  

NetCDF support is not implemented. Includes one test.

related to #480

### DIFF
--- a/src/ewatercycle/observation/grdc.py
+++ b/src/ewatercycle/observation/grdc.py
@@ -395,6 +395,10 @@ def get_grdc_data_monthly(
     Returns:
         grdc data in a xarray dataset.
             Shaped like a filtered version of the GRDC daily NetCDF file.
+        Data description:
+            Original   - original (provided) data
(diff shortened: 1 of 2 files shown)


andrevdv:
as for the `unit_of_measure`: 

I kept using the the hardcoded `units": "m3/s"` for both `Original discharge` and `Calculated discharge` similar to how it's done for daily data so the metadata field for `unit_of_measure` is still unused. Could add it to the attributes list for completeness.

andrevdv:
using actual GRDC `.txt` file:

example

MarkMelotto (approved):

MarkMelotto:
@andrevdv looking good, good job!
The only thing is this testing thing failing, not sure what is going on...

PR #495, opened 2026-02-02
Update documentation for CARAVAN_CACHE variable

MarkMelotto:
Clarify the environmental variable path for CARAVAN_CACHE.

It is nice to know how to actually use it :)

### DIFF
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -112,9 +112,9 @@ class CaravanForcing(DefaultForcing):
         By default, it will open the dataset from data.4tu.nl OPeNDAP server
         This can be overridden by having an environmental variable: CARAVAN_CACHE.
         Set this variable to the directory containing the netCDF files.
-        On an eWaterCycle machine this path is: /data/shared/climate-data/caravan_data.
+        On an eWaterCycle machine this path is: /data/shared/climate-data/caravan.
         On other machines it depends on how you mount the dcache folder.
-        It should end with /caravan_data.
+        For example you mount /climate-data/caravan to home/caravan_data.
         cli: export CARAVAN_CACHE=/some/dcache/path/caravan_data
 
         Args:
--- a/src/ewatercycle/_forcings/caravan.py
+++ b/src/ewatercycle/_forcings/caravan.py
@@ -114,9 +114,8 @@ class CaravanForcing(DefaultForcing):
         Set this variable to the directory containing the netCDF files.
         On an eWaterCycle machine this path is: /data/shared/climate-data/caravan_data.
         On other machines it depends on how you mount the dcache folder.
-        It should end with /caravan_data
-        
-
+        It should end with /caravan_data.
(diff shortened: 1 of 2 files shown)


MarkMelotto:
@BSchilperoort when you have time, please take a look, should not take more than a minute!

src/ewatercycle/_forcings/caravan.py
```diff
@@ -112,7 +112,10 @@ def get_dataset(cls: type["CaravanForcing"], dataset: str) -> xr.Dataset:
         By default, it will open the dataset from data.4tu.nl OPeNDAP server
         This can be overridden by having an environmental variable: CARAVAN_CACHE.
         Set this variable to the directory containing the netCDF files.
-
+        On an eWaterCycle machine this path is: /data/shared/climate-data/caravan.
+        On other machines it depends on how you mount the dcache folder.
+        For example you mount /climate-data/caravan to /path/to/caravan_data.
+        cli: export CARAVAN_CACHE=/path/to/caravan_data
```
```suggestion
        On other machines it depends on where the data is located.
        If you have created a local copy of the data at /path/to/caravan_data,
        use the following command in your terminal;
            export CARAVAN_CACHE=/path/to/caravan_data
        To always have this environment variable set, add this to .bashrc or
        ask your system administrator. 
```

BSchilperoort (changes requested):
This is a more generic way to put it I think. Not every ewatercycle setup needs to have dcache, it's just what we use because it's the most convenient thing SURF offers for this.

src/ewatercycle/_forcings/caravan.py
MarkMelotto: would you still call it a local copy, even though it points to a webdav mount outside of your environment?
I think that will be unclear to users who do not know what mounting is.

BSchilperoort: I guess we can say "locally available copy" to make it a bit more generic. A mount still appears as "local" in your filesystem.

> even though it points to a webdav mount outside of your environment

It doesn't need to be. That's where you have stored it. Users can also just store it on a local drive, or if they're on snellius/other HPC on `/scratch-shared/` or something.

MarkMelotto: Yeah I think that works better, as not all users will know "mounting" and you are definitely right on the storing of the data on a local drive!

BSchilperoort (approved):

MarkMelotto:
@BSchilperoort there is this new rule that wants to change lines like:
```dataset: str = basin_id.split("_")[0]```

to:
```
parts = basin_id.split("_")
dataset = parts[0]
```

I think that makes readability worse, so quick check if you agree

BSchilperoort:
> @BSchilperoort there is this new rule that wants to change lines like: `dataset: str = basin_id.split("_")[0]`
> 
> to:
> 
> ```
> parts = basin_id.split("_")
> dataset = parts[0]
> ```
> 
> I think that makes readability worse, so quick check if you agree

I agree!

PR #500, opened 2026-09-03
eWaterCycle upgrade to V2.5

MarkMelotto:
Testing and changing small things to be ready for a working 2.5 version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,43 @@ Formatted as described on [https://keepachangelog.com](https://keepachangelog.co
 
 ## Unreleased
 
-## Added
+## [2.5.0] (2026-09-07)
+
+### Added
 
 - `model.logs` property to inspect the logs of a containerized model ([#470](https://github.com/eWaterCycle/ewatercycle/pull/470)).
-- mention remotebmi in docs ([#471](https://github.com/eWaterCycle/ewatercycle/issues/471))
+- `ewatercycle.util.plot_catchment()` to plot a catchment shapefile, either as a new figure or onto an existing cartopy `GeoAxes` ([#485](https://github.com/eWaterCycle/ewatercycle/pull/485)).
+- `ewatercycle.observation.grdc.get_grdc_data_monthly()` for reading monthly GRDC station files (`.txt` only; NetCDF is not supported). Returns original streamflow, GRDC-calculated streamflow and the validity flag ([#494](https://github.com/eWaterCycle/ewatercycle/pull/494)).
+- `hydrograph()` now accepts `xarray.Dataset` and `xarray.DataArray` in addition to `pandas.DataFrame` ([#493](https://github.com/eWaterCycle/ewatercycle/pull/493)).
+- `metrics_list` argument to `hydrograph()`, to choose which goodness-of-fit metrics are shown. Metrics can be given by full name, abbreviation or function name (any [HydroErr metric](https://hydroerr.readthedocs.io/en/stable/list_of_metrics.html)), or as a callable ([#493](https://github.com/eWaterCycle/ewatercycle/pull/493)).
+- `selected_year` argument to `hydrograph()`, to plot a single year from a longer time series ([#493](https://github.com/eWaterCycle/ewatercycle/pull/493)).
+- `ERA5-Land` to the predefined forcing datasets (`ewatercycle.forcing.DATASETS`) ([#499](https://github.com/eWaterCycle/ewatercycle/pull/499)).
+- `CARAVAN_CACHE` environment variable. When set to a directory containing the Caravan NetCDF files and a `shapefiles/combined.shp`, `CaravanForcing` reads from that local copy instead of the 4TU OPeNDAP server. On eWaterCycle machines Caravan v1 is at `/data/shared/climate-data/caravan` and Caravan v1.6 at `/data/shared/climate-data/caravan1_6` ([#492](https://github.com/eWaterCycle/ewatercycle/pull/492), [#495](https://github.com/eWaterCycle/ewatercycle/pull/495)).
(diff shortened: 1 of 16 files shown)


MarkMelotto:
@andrevdv please add yourself to the .cff file :)

MarkMelotto:
@BSchilperoort I have a question regarding numpy. ESMValCore uses version 1.26.4 right now and is locked. But in this pyproject.toml what do we use? because it will override the numpy version for HydroErr?

BSchilperoort:
> @BSchilperoort I have a question regarding numpy. ESMValCore uses version 1.26.4 right now and is locked. But in this pyproject.toml what do we use? because it will override the numpy version for HydroErr?

I think the only way to solve this is to update all dependencies to new versions and fix what breaks. Would be good to do anyway (more modern python, new esmvaltool/core version that soon supports zarr)

MarkMelotto:
> I think the only way to solve this is to update all dependencies to new versions and fix what breaks. Would be good to do anyway (more modern python, new esmvaltool/core version that soon supports zarr)

I think it is best that that will happen in a new branch/PR 'getting ready for release 3.0' or something like that.
I am then happy with this 2.5 version as it is now.

BSchilperoort:
> I think it is best that that will happen in a new branch/PR 'getting ready for release 3.0' or something like that.
> I am then happy with this 2.5 version as it is now.

OK. would be fine as a 2.6 release as well though. Nothing important should break

pyproject.toml
BSchilperoort: Hydroerr is pinned to a specifc version. This is bad practice. Would be better to do `>=2.0.0` or  `=~2`.

CHANGELOG.md
```diff
@@ -6,10 +6,43 @@ Formatted as described on [https://keepachangelog.com](https://keepachangelog.co
...
+
+- Makkink forcing generation now works with any climate model dataset. CF decoding is deferred until after loading (keeping Dask happy), `time_bnds` is loaded into memory before merging, and the half-day time offset uses `pandas.Timedelta` instead of `numpy.timedelta64` ([#487](https://github.com/eWaterCycle/ewatercycle/pull/487)).
+- The half-day offset in `merge_esvmaltool_datasets()` uses the `"12h"` frequency alias instead of the deprecated `"12H"`, which raises a `FutureWarning` on pandas 2.2 and is removed in pandas 3.0 ([#500](https://github.com/eWaterCycle/ewatercycle/pull/500)).
+- Monthly GRDC files could not be read, because they use a day of `00` that pandas cannot parse. These dates are now mapped onto the first day of the month ([#500](https://github.com/eWaterCycle/ewatercycle/pull/500)).
+- Selecting a Caravan basin now works whether the basin IDs are stored as bytes or as strings ([#500](https://github.com/eWaterCycle/ewatercycle/pull/500)).
+- Broken user guide link in the README ([#474](https://github.com/eWaterCycle/ewatercycle/pull/474)).
+
+### Dependencies
+
+- Added `HydroErr==2.0.0`.
```
BSchilperoort: see comment on pyproject.toml file

BSchilperoort (approved):
Please change the hydroerr version specification before merging.

Don't forget to update the conda lockfile at ewatercycle/infra.

MarkMelotto:
> Don't forget to update the conda lockfile at ewatercycle/infra.

Isnt the lockfile here in ewatercycle/ewatercycle?

Oh you mean this one: https://github.com/eWaterCycle/infra/blob/main/roles/ewatercycle/defaults/main.yml?

BSchilperoort:
> > Don't forget to update the conda lockfile at ewatercycle/infra.
> 
> Isnt the lockfile here in ewatercycle/ewatercycle?
> 
> Oh you mean this one: https://github.com/eWaterCycle/infra/blob/main/roles/ewatercycle/defaults/main.yml?

Oh sorry the lock file is in this repo. Needs updating after changing dependencies. But that file should also be updated.

MarkMelotto:
> Oh sorry the lock file is in this repo. Needs updating after changing dependencies. But that file should also be updated.

I am thinking now of leaving the lockfile as is, only hydrostats was removed
Because of this line: 
[tool.conda-lock.dependencies]
esmvaltool-python = ">=2.11"

could simply edit it...

OR

I indeed move up to ESMValTool 2.15, which may take more time than I can currently spend on it.
So is it fine to keep it as is and then later we update ESMValCore etc?

That does leave the main.yml that needs to change

BSchilperoort:
> I am thinking now of leaving the lockfile as is, only hydrostats was removed

~~The [lockfile is used to setup ewatercycle on SRC](https://github.com/search?q=repo%3AeWaterCycle%2Finfra+conda-lock.yml&type=code). So 2.5 would not be available there out-of-the-box. The lockfile has hydroerr but it's <2~~

~~Updating the lock file should not mean you need to move to a newer esmvaltool version. Just to generate a lockfile that works for ewc 2.5.0~~

**Edit:** not a 100% sure it's needed actually...

MarkMelotto:
> > I am thinking now of leaving the lockfile as is, only hydrostats was removed
> 
> ~The [lockfile is used to setup ewatercycle on SRC](https://github.com/search?q=repo%3AeWaterCycle%2Finfra+conda-lock.yml&type=code). So 2.5 would not be available there out-of-the-box. The lockfile has hydroerr but it's <2~
> 
> ~Updating the lock file should not mean you need to move to a newer esmvaltool version. Just to generate a lockfile that works for ewc 2.5.0~
> 
> **Edit:** not a 100% sure it's needed actually...

I think just updating the main.yml in infra is fine for now, awaiting your response.
(I am not that familiar with lockfiles, im glad I read that they are built, I was going to add things manually...)

BSchilperoort:
> I think just updating the main.yml in infra is fine for now, awaiting your response.

I think that will be OK. Best to test creating a new machine on SRC once the release is out

MarkMelotto:
> > I think just updating the main.yml in infra is fine for now, awaiting your response.
> 
> I think that will be OK. Best to test creating a new machine on SRC once the release is out

that was definitely on my todo list already 😄

PR #502, opened 2026-09-10
Bumping esmvaltool to 2.15.* giving us: eWC v2.6

MarkMelotto:
Updated the esmvaltool to version 2.15.*.
This changed the conda-lockfile which was then tested and small bugs were fixed.

Making this ready for eWaterCycle 2.6. (now not doing the release myself ;))

In [ewatercycle/infra](https://github.com/eWaterCycle/infra/pull/187) there is also a v2.6 version ready with changes to conda environments to keep up with esmvaltool changes. When that one is accepted I will add and test the [DestinE plugin](https://github.com/eWaterCycle/ewatercycle-destine).

On the DestinE plugin: it now will give DestineELumpedMakkinKForcing etc, but they are done in a way that works now. For future reference that should become LumpedMakkinkForcing and in the generate there should be the option to select DestinE data.


pyproject.toml
```diff
@@ -105,7 +103,8 @@ platforms = [
 ]
 
 [tool.conda-lock.dependencies]
-esmvaltool-python = ">=2.11"
+python = "3.12.*"
+esmvaltool-python = "2.15.*"
```
BSchilperoort:
```suggestion
esmvaltool-python = ">=2.15"
```
pinning this to a specific version here isn't ideal; it will make it more difficult for others to upgrade the esmvaltool version later (as pip will complain that it's incompatible with ewatercycle)

src/ewatercycle/testing/data.py
```diff
@@ -0,0 +1,27 @@
+"""Data files shipped with ewatercycle for testing and examples.
+
+Kept separate from :mod:`ewatercycle.testing.fixtures` because that module
```
BSchilperoort, line 3: I don't think this is the completely correct way to fix this. Pytest is not a package dependency, so `import pytest` should not have been in any code in `src/`.

For completeness I would also move the pytest fixtures (in `src/ewatercycle/testing/fixtures.py`) to `tests/`, and in the tests fixtures import the `rhine_shape` function from this file.

tests/src/baseline_images/test_analysis/hydrograph_DataFrame.png
BSchilperoort, line 1: I don't think these images should be (or have been) comitted to the repository. They are not used for validation.

Might be OK to keep them as they were originally, but these minor visual changes are not worth adding another MB to the git history.
You will have to revert the changed images in this PR and do a squash merge.

BSchilperoort (changes requested):
Nice work, I assume you tested it sufficiently on SRC. I currently do not have the time to test it myself.

I found two small issues which should be improved (shouldn't take too much time).

src/ewatercycle/testing/data.py
MarkMelotto: that is indeed a cleaner fix, thanks for your insight!

MarkMelotto:
> Nice work, I assume you tested it sufficiently on SRC. I currently do not have the time to test it myself.

Yes, everything seemed to work as intended with generating Forcings.

> I found two small issues which should be improved (shouldn't take too much time).

On it, thank you for taking the time to review!

tests/src/test_analysis_hydrograph.py
BSchilperoort, line 1: Would be maybe better to still save them but add a .gitignore file in the directory to not have people commit the images accidentally.

BSchilperoort (approved):
Small comment.

Don't forget to squash merge to not have the image files in git history :)

tests/src/test_analysis_hydrograph.py
MarkMelotto: We can also just remove them right?

Otherwise I will restore the saving and indeed add them to the .gitignore

MarkMelotto:
> squash merge

you mean just in the PR right? or something else?

I am first checking and testing if everything works

BSchilperoort:
> you mean just in the PR right? or something else?

Yeah here. 

image

Seems like it's the default and no option to deviate haha. So that'll be OK

tests/src/test_analysis_hydrograph.py
BSchilperoort: The nice thing about the saving of figures is that they can be visually assessed. Now that's not possible (if figs aren't saved).

MarkMelotto: alright then I am reverting and adding them to the gitignore

docs/user_guide/00_intro.ipynb
BSchilperoort, line 1: These changes make the docs hard to read. Would be better to not show all parameter sets in the config

BSchilperoort: rich's pprint has a max_depth argument; https://rich.readthedocs.io/en/latest/reference/pretty.html#rich.pretty.pprint

docs/user_guide/03_models_obs_analysis.ipynb
BSchilperoort, line 1: Too many lines I think here.

print(
        model_instance.time_as_isostr,

prints every model timestep and it's not wiped here by \r

MarkMelotto: yep already on it :)

docs/user_guide/00_intro.ipynb
MarkMelotto: then leave them in 01_parameter_sets?

MarkMelotto:
@BSchilperoort I think I got it now, rich pprint makes me smile, amazing solution!
I will leave the merge for tomorrow so I can fix unseen issues!

Thanks for the help and time you put in!

BSchilperoort:
> @BSchilperoort I think I got it now, rich pprint makes me smile, amazing solution! I will leave the merge for tomorrow so I can fix unseen issues!
> 
> Thanks for the help and time you put in!

No problem. Do note the notebook 01_parameter_sets has a similar issue with a big list of parameter sets.

MarkMelotto:
> No problem. Do note the notebook 01_parameter_sets has a similar issue with a big list of parameter sets.

I was thinking: keep it that way to show what is already there

BSchilperoort:
> I was thinking: keep it that way to show what is already there

It does make the [documentation page](https://ewatercycle--502.org.readthedocs.build/en/502/user_guide/01_parameter_sets.html) quite unreadable. And the full list is very specific to which infrastructure it's run on. We hope that in the future eWaterCycle will have users outside TU Delft/the Netherlands. :)

MarkMelotto:
> > I was thinking: keep it that way to show what is already there
> 
> It does make the [documentation page](https://ewatercycle--502.org.readthedocs.build/en/502/user_guide/01_parameter_sets.html) quite unreadable. And the full list is very specific to which infrastructure it's run on. We hope that in the future eWaterCycle will have users outside TU Delft/the Netherlands. :)

Got it, that does indeed make sense! Changed it now :)

